Issue #42, opened 2020-04-01
installation


jameshunterbr:
I have tried a number of times to install vizR on Windows, but it keeps failing. The latest failure has to do with the files/packages (?) RTCGA and RTCGA.clinical as below. What am I doing wrong or what is the bug involved?
Thanks.
James Hunter

```
√  checking for file 'C:\Users\james\AppData\Local\Temp\Rtmp8McCI4\remotes77f838e42777\openpharma-visR-0c1b345/DESCRIPTION' (893ms)
-  preparing 'visR': (557ms)
√  checking DESCRIPTION meta-information ... 
-  excluding invalid files
   Subdirectory 'man' contains invalid file names:
     'visR architecture.png'
-  checking for LF line-endings in source and make files and shell scripts
-  checking for empty or unneeded directories
-  building 'visR_0.1.0.tar.gz'
   
Installing package into ‘C:/Users/james/OneDrive/Documents/R/win-library/3.6’
(as ‘lib’ is unspecified)
ERROR: dependencies 'RTCGA', 'RTCGA.clinical' are not available for package 'visR'
* removing 'C:/Users/james/OneDrive/Documents/R/win-library/3.6/visR'
Error: Failed to install 'visR' from GitHub:
  (converted from warning) installation of package ‘C:/Users/james/AppData/Local/Temp/Rtmp8McCI4/file77f8328a59f6/visR_0.1.0.tar.gz’ had non-zero exit status
```

1beb:
You need to install the dependencies.

```
install.packages(c("RTCGA", "RTCGA.clinical") 
```

Alternatively you can use remotes with the argument `dependencies=TRUE`: 

```
remotes::install_github("openpharma/visR", dependencies=TRUE) 
```

Kawameicha:
RTCGA and RTCGA.clinical are both available from Bioconductor (an open source project that hosts a wide range of tools for analyzing biological data with R).

First, you'll need Bioconductor itself:

`if (!requireNamespace("BiocManager", quietly = TRUE))
    install.packages("BiocManager")
BiocManager::install(version = "3.10")`

Then, of course the two packages:

`BiocManager::install(c("RTCGA", "RTCGA.clinical"))`

After what you should be able to install vizR on your machine.

jameshunterbr:
Thanks. Got them and visR downloaded perfectly after that. You may want to update your read.me file to help folks avoid this issue. 

Best, 

Jim Hunter 

⁣Get BlueMail for Android ​

PR #44, opened 2020-04-02
Add `Remotes` for Bioconductor packages

thomas-neitmann:
This fixes the installation issue describes in #42. To automatically install packages from Bioconductor the `Remotes` field in `DESCRIPTION` has to be filled.


cschaerfe:
We can probably take out TCGA dependency since this was only needed for an older version of the vignette illustrating TTE analysis using TCGA data. Thanks for fixing - I'll merge for now so install of the current version works.

Issue #26, opened 2020-03-25
Set up CI/CD workflow


bailliem:
Post webinar set up a CI/CD workflow including
* unit tests
* dependencies 
* styling and linting
* documentation build
* website build

kawap:
I could take this task and prepare CI/CD workflow in Travis .. but I am wondering if this possible to use Travis CI in Roche-phcaa for usage with github, not github.roche?

bailliem:
That would be great if you were happy to look in to it. If there was a workflow that could allow this to work with cross organisation collaboration, etc.

PR #49, opened 2020-04-15
GitHub Actions CI

kawap:
PR to issue ref: #26 

Finally I have used Github Actions to set up CI workflow. I have set up basic CI workflow which is evaluating pipelines in linux and windows environment. Moreover actions/cache makes that CI workflow is much more faster (caching dependencies) 

Working on that I have met some troubles. During CI build, after installation of all newest versions of dependencies, Alluvial vignette throw mysterious error while running vr_alluvial_plot() function:
`The subscript 'tibble_vars[[i]]' has value 0 but must be a positive location.`.
I have been digging and It takes me a piece of time to find that it is as follows: 
function `easyalluvial::alluvial_long()` is calling inside` dplyr::select_if()` on tibble handling data where columns in the tibble are named using 'i' (iterator).  In the past there have been already problem like that: https://github.com/tidyverse/dplyr/issues/4380 but this time I have found that it is not caused by updating of dplyr but latest tibble 3.0.0 released 16 days ago. When I force installation of previous version of  tibble in DESCRIPTION FILE then everything works fine. So it is temporary solution until tibble package team will not fix that.  

Also for some reason there has been problem with automate installing of : "rstantools", "RBesT", "rcmdcheck" and I decided to add a line with 'manual' installation before installing all dependencies what solved the case.  

For now package is build including vignettes and check throw:
0 errors ✔ | 8 warnings ✖ | 8 notes ✖
**so we have 0 errors but warnings need to be fixed.**

details are here: https://github.com/openpharma/visR/runs/587378443?check_suite_focus=true

For now CI build on GitHub Actions CI branch have failed only because I have set restrictive option : `error_on = "warning"` . It is option what we need to discuss of course.  

In next step I will add CD part of workflow (rebuilding of docs file).


kawap:
Latest commits introduce Continuous Deployment part of CI/CD workflow and made CI/CD build status badge in README file.

In result, workflow is set as follows:

**CI_CD.yaml** is triggered only by push actions in branches **master and develop** and contain:

- [ ] Install dependencies

- [ ] `rcmdcheck()`   (error_on = "warning" will be enabled after improve of warnings by @thomas-neitmann )

- [ ] Render README

- [ ] Build visR package website using `pkgdown::build_site()` and push docs files to branch whose change triggered CI/CD workflow 

**R-CMD-check** workflow is triggered  in **other branches**  and contain:

- [ ] Install dependencies

- [ ] `rcmdcheck()`   (error_on = "warning" will be enabled after improve of warnings by @thomas-neitmann )

kawap:
image

PR #61, opened 2020-05-21
Unstratified estimates, summary and plot.

SHAESEN2:
Allow for unstratified estimates, summary and plot by adding $strata to list.


epijim:
@SHAESEN2 - what was the plan on the removal of the master wrapper? I can see a reason I think (steps are pretty easier - and you want to do things on the object - but it means the vignette's don't work very well now)

PR #67, opened 2020-09-20
Merge sandbox

bailliem:
Merge visR-survival repository and maintain commit history https://github.com/visR-sandbox/visR-survival.
* Experimental api for survival time2event plotm including risk table and table summaries.


DESCRIPTION
```diff
@@ -25,23 +28,22 @@ Suggests:
...
     broom,
+    RColorBrewer,
+    DT,
+    survival,
     magrittr,
     ggplot2,
     patchwork,
-    tibble,    
+    tibble,
     RBesT,
```
epijim, line 47: Maybe something to tackle later - but we have some dependencies that maybe aren't essential. e.g. for something like`RBest` which is a bit painful, maybe we could save just the object rather than generating it for unit testing/docs.

DESCRIPTION
```diff
@@ -54,5 +56,14 @@ Imports:
...
+    easyalluvial (>= 0.2.2),
+    cowplot,
+    gtable,
+    here,
+    gridExtra,
+    plotly,
+    reshape2,
+    rlang,
+    base,
+    openxlsx
```
epijim, line 68: Same comment about dependencies for openxlsx

epijim (approved):

PR #87, opened 2020-10-13
Rbest remove from depends 

bailliem:
This is an experimental function and needs to be developed. At the moment this dependency is causing errors in the visR package. This has been removed and the tidy rbest function moved to indicate it is experimental.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -44,7 +44,6 @@ Imports:
     ggplot2,
     patchwork,
     tibble,
-    RBesT,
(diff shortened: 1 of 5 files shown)


bailliem:
@SHAESEN2  and @cschaerfe I've removed the dependency with Rbest to avoid any issues with the visR install. I've kept the functions in R, but added badges to the function descriptions to indicate they are experimental.

SHAESEN2 (approved):
Looks great.

Issue #97, opened 2020-11-18
Filtering in attrition returning inconsistent results


cschaerfe:
#### Issue description
There seems to be an issue with the attrition code when using logical operators in the filtering strings. The table does not return correct N's

#### Steps to reproduce the issue

Please share code if possible.

```r
data(lung)
lung_cohort <- lung %>% 
    dplyr::rename(ECOG = ph.ecog,
                  Karnofsky = ph.karno,
                  SITEID = inst,
                  CNSR = status,
                  AVAL = time,
                  WEIGHT = wt.loss
                 ) %>% 
    dplyr::mutate(USUBJID = sprintf("Pat%03d", row_number()),
                  SEX = factor(if_else(sex == 1, "male", "female")),
                  CNSR = if_else(CNSR == 2, 1, 0),
                  SUBGR1 = factor(case_when(ECOG == 0 ~ "0 asymptomatic",
                                            ECOG == 1 ~ "1 ambulatory",
                                            ECOG == 2 ~ "2 in bed less than 50% of day",
                                            ECOG == 3 ~ "3 in bed more than 50% of day",
                                            ECOG == 4 ~ "4 bedbound",
                                            ECOG == 5 ~ "5 dead")
                                 ),
                  AGEGR1 = factor(case_when(age < 30 ~ "< 30y",
                                            age >= 30 & age <= 50 ~ "30-50y",
                                            age > 50 & age <= 70 ~ "51-70y",
                                            age > 70 ~ "> 70y"),
                                  levels=c("< 30y", "30-50y", "51-70y", "> 70y")
                                 )
                 ) 
cohort_attrition <- vr_attrition_table(
  data = lung_cohort,
  criteria_descriptions = c("1. ECOG available", 
                            "2. Weight loss data available", 
                            "3. Non-missing censoring status",
                            "4. Positive follow up time",
                            "5. Meal Cal (missing or  > 1000"),
  criteria_conditions   = c("!is.na(SUBGR1)",
                            "!is.na(WEIGHT)",
                            "!is.na(CNSR)",
                            "AVAL >= 0",
                            "is.na(meal.cal) | meal.cal >= 1000"),
  subject_column_name   = 'USUBJID')
```
returns 132 patients after all filtering.

#### What's the expected result?

```
lung_cohort %>%
  dplyr::filter(!is.na(SUBGR1),
                !is.na(WEIGHT),
                !is.na(CNSR),
                 AVAL >= 0) %>% 
[...]

PR #98, opened 2020-11-18
Fixed issue #97. Attrition table not calculated properly

rebecca-albrecht:
due to an error in operator precedence. Specifically, conditions where added to a string in the form "c1 & c2" but single conditions were not put into brackets. In a condition included another binary operator, e.g. c1 = a1 | a2, then the string would be a1 | a2 & c2. Now it should correctly be (a1 | a2) & c2.

Fixes https://github.com/openpharma/visR/issues/97


cschaerfe (commented):
Looks good to me

PR #112, opened 2021-03-03
Added some schematics for test / tests for the get_XXX functions

rebecca-albrecht:


cschaerfe (approved):

SHAESEN2 (commented):
It is better to finish tests one function at a time than provide templates. With templates it is difficult to follow what is finished and what not.

PR #116, opened 2021-03-04
Modified function to not throw a warning when using default parameter…

dazim:
…s without changing plotting behaviour

### DIFF
--- a/R/add_CI.R
+++ b/R/add_CI.R
@@ -40,7 +40,7 @@ add_CI <- function(gg, ...){
 add_CI.ggsurvfit <- function(gg, 
                              alpha = 0.1, 
                              style = "ribbon", 
-                             linetype = "2a", ...){
+                             linetype, ...){
   
   if (! base::all(c("est.lower", "est.upper") %in% colnames(gg$data))) {
     warning("Confidence limits were not part of original estimation.")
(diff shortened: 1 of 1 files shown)


R/add_CI.R
```diff
@@ -37,31 +37,53 @@ add_CI <- function(gg, ...){
...
+                                        ymax = est.upper, 
+                                        fill = strata), 
+                           alpha = alpha, 
+                           na.rm = TRUE)
   }
-
+  
   if (style == "step"){
+    
+    if (linetype == "2a") {
```
SHAESEN2, line 66: I dont understand the issue here.

R/add_CI.R
```diff
@@ -37,31 +37,53 @@ add_CI <- function(gg, ...){
 #' @method add_CI ggsurvfit
 #' @export
 
-add_CI.ggsurvfit <- function(gg, alpha = 0.1, style = "ribbon", linetype = 2, ...){
-
+add_CI.ggsurvfit <- function(gg, 
+                             alpha = 0.1, 
+                             style = "ribbon", 
+                             linetype = "2a", ...){
```
SHAESEN2, line 43: Linetype "2a" is not a sensible default,, neither is a hack.  Please update. I dont understand the issue. Perhaps we can discuss this to look for a solution? I never had an issue with this function honestly.

SHAESEN2 (changes requested):

R/add_CI.R
dazim: There isn't really an "issue", but calling the function with its default parameters yields a warning. Rewriting it like this doesn't change the behavior as presented to the user but doesn't cause a warning. In my opinion, the default behavior of shouldn't automatically cause a warning since that'd motivate (at least me) to investigate this further.

dazim: image
This is the current behavior

dazim: Without giving any further parameters, it defaults to `style = "ribbon"` and `linetype = 2`. Right now, this triggers the warning in line 53 (`warning("Argument linetype not used for style ribbon")`) which the user per se did not cause. This would irritate me personally. I've now pushed a less hacky looking fix that now only prints a warning if the user manually specified a linetype while using ribbon and sets it to solid when using step.

PR #118, opened 2021-03-05
Added tests for get_quantiles.R

timtreis:
Testing standard input.
Testing generated quantiles against an equivalent call using pure survival package.


SHAESEN2 (changes requested):
We still need to write get quantiles and we don't expect them to be the same as the ones obtained by R survival functions. Let us discuss before full blown testing.

SHAESEN2 (commented):
I am adding tests in the get_quantiles branch.
These test don't cover the function well enough
Use the function name instead of plot

PR #124, opened 2021-03-18
Added vdiffr for visual testing of add_CI

dazim:
Also includes extremely minor change to the wording of add_CI's @param.


bailliem (commented):
@dazim - Looks ok. Is the Rplots.pdf required. Looks like a stray file.

bailliem:
Lets keep Rplots.pdf. I'll leaved it as a todo, if there is a requirement from vdiffr which generates it.

timtreis:
I restored the branch since I wanted to use it for testing the other add_X functions, just so that you know why I restored it :D

PR #125, opened 2021-03-18
Review documentation add_annotation 

bailliem:
* updated example to reflect change to api vr_plot to visR::plot()
* added additional example to display a string comment


dazim:
Works and explains it well. Do we want to include a primitive example for annotations outside of the plotting area?

SHAESEN2 (approved):

timtreis (changes requested):
I think we should be coherent in our example in the usage of package::function. Therefore I suggest the following changes:
- Line 29 estimate_KM() -> visR::estimate_KM
- Line 48 get_pvalue() ->  visR::get_pvalue()

Also, for this particular example, it is only necessary to explicitly load `visR` and `survival`. The others are loaded by proxy.

bailliem:
@timtreis  - Can you also check you agree we can remove library(survival) in the examples

timtreis (approved):
Verified to works on my machine without importing survival. Merge would be fine for me.

bailliem:
Closed merge request.

PR #126, opened 2021-03-18
Added missing tests for warning/error behaviour. Should raise test coverage to 100% for CI and CNSR.

timtreis:
Also improved our warn/test behaviour in add_CI.R to make more sense.


bailliem:
TODO: Add information within pipe to catch errors

PR #132, opened 2021-03-18
Documentation for plot function

bailliem:
Add examples to illustrate that visR::plot overrides base plot when visR loads.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,7 +52,6 @@ Imports:
     tidyr,
     dplyr,
     ggpubr,
-    ggsci,
     pammtools,
     kableExtra,
     assertthat,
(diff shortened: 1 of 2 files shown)


bailliem:
TBD - renaming plot to avoid overriding base plot

timtreis (changes requested):
Hope I'm not being annoying with this :D

- Line 46 -> survival::survfit
- Line 58 -> visR::estimate_KM

SHAESEN2 (approved):
If we document as such, I am fine with keeping visR::plot

timtreis (approved):

bailliem (commented):
Agree to keep as is, but check experience on possible users

PR #135, opened 2021-03-19
Correct weird behavior survival with only one stratum

SHAESEN2:


timtreis (approved):
Corrects behaviour.
Manually verified tests. 
Works on my machine (different OS I guess? Ubuntu 64 bit).

PR #143, opened 2021-03-19
Test/add pvalue

SHAESEN2:
Rewrote get_pvalue and include associated tests.


timtreis:
Just a quick note: testthat suggests to put helper functions under R/

https://testthat.r-lib.org/reference/test_dir.html#special-files
https://www.r-bloggers.com/2020/11/helper-code-and-files-for-your-testthat-tests/

rebecca-albrecht (approved):

SHAESEN2:
Hey Tim, based on those links it is OK and recommended to keep them where they are. Putting it in R/ will raise questions as there is no documentation etc.

timtreis:
Do they? In the end I don't really care as long as we don't have them in both places. I just understood the following part as "put them under R/"

> There are two other types of special file that we no longer recommend using:
> 
> * Helper files start with helper and are executed before tests are run. They're also loaded by devtools::load_all(), so there's no real point to them and you should just put your helper code in R/.

SHAESEN2:
If we follow this advice, the package does not want to install. I'll keep as is.

timtreis:
Okay, then I'll move my helper functions to the testthat folder as well :)

PR #152, opened 2021-04-16
Replaced parameter so that no error is caused by ggplot

timtreis:
To fix the vdiffr error while checking


bailliem:
@timtreis - that should be the trick :-)

PR #154, opened 2021-04-17
Fix/cox

SHAESEN2:


rebecca-albrecht (approved):
Works fine, thank you!

bailliem (approved):
Looks fine. I noted one thing I'll put as a todo for me, is to check how we manage the dependency with Survminer and ggsurvfit in the description file. Survminer was not automatically loaded on install and a note was given.

PR #159, opened 2021-04-21
Test/tidyme

SHAESEN2:


timtreis (changes requested):
There is still the remainder of debug output in line 97 :wink:  Other than that, it can be merged from my end. Remember to not delete the branch since we can't reuse names AFAIK

PR #166, opened 2021-05-07
Feature/attrition

cschaerfe:
Added attrition features with new API including `get_attritition`, `apply_attrition` and `plot`. Tests are provided for get and apply, but not yet for plotting.


bailliem:
@cschaerfe  -- I've resolved the merge conflicts. There was an update to plot.R that was not reflected in your branch.

cschaerfe:
Testing helps! Found one more bug that I fixed in cases where one would use a concatenated filter using OR (e.g., "AGE > 61 | SEX == 'M'" followed by additional filters

bailliem (approved):

DESCRIPTION
```diff
@@ -33,8 +33,7 @@ Suggests:
     forcats,
     glue,
     here,
-    stringr,
-    dplyr
+    stringr
```
SHAESEN2, line 36: stringr is a package that we don't really need. These are wrappers around base sub gsub and trimws

SHAESEN2 (approved):
Do we need stringr?

DESCRIPTION
cschaerfe: Good point. I think I used it, but we can consider taking it out again, if we want to

cschaerfe: It only shows up under suggests, so I guess we can keep it here and then clean up description before the release?

rebecca-albrecht (approved):

PR #171, opened 2021-05-09
Resolve plot warning rmd check

bailliem:
1. change survfit to survfit_visR class in estimate_KM. This is extension of survfit
2. remove rewriting default plot within package. just export s3 plot class
3. Change plot to plot.survfit_visR to expect survfit_visR object now.
4. survival::survfit.plot is not overrided. so plot(surfit) is original survival plot, while plot(survfit_visr) is visR plots. 
5. remove all reference to visR::plot() as plot is a generic s3 of graphics::plot which expects survfit_visR object


SHAESEN2 (changes requested):
Not in favor of explicitly changing the class. We will lose the the advantage that our plotting function can take any survival object. That is one of the things why I dont use survminer. You need a survminer object for everything to work.

bailliem:
@SHAESEN2 - I agree with you. How do you feel about changing the plot function name to plot_vr or equivalent?

cschaerfe:
I am also in favor of changing the name of the plotting function instead of the object class.

bailliem:
I think we are moving towards renaming. I'll close this pull request, and open up a new one with the renaming when we decide on it.

bailliem:
Closing pr. Decision to rename plot function

PR #177, opened 2021-05-11
Rename plot function

bailliem:
* Rename plot to visR
* Replace all instances of visR::plot with visR::visr
* rename plot.R to plot_visr.R
* tidy up documentation


SHAESEN2 (changes requested):
If we are going to be consistent I would also change:
- utils_plot.R => utils_visr.R

R/visr_plot.R
```diff
@@ -101,13 +109,13 @@ plot.survfit <- function(
       log = "log(Survival probability)",
       event = "Failure probability",
       cloglog = "log(-log(Survival probability))",
-      pct = "Survival probability (%)",
-      logpct = "log(Survival probability (%))",
+      pct = "Survival percentage (%)",
```
SHAESEN2, line 112: Not sure if this is the correct term.

SHAESEN2 (approved):
Not sure if  'survival percentage' is the correct wording. Perhaps Percentage survival might be better?

R/visr_plot.R
bailliem: Agree. Survival percentage makes more sense. Also we can then drop the %.

bailliem: Same for log surv percentage

SHAESEN2: Did you mean Percentage survival makes more sense? (survival percentage is what it is saying now)

bailliem: Percentage survival.

SHAESEN2:
Still some weird phrasing present. Cloglog is not in percentage and you forgot to change one survival percentage

bailliem:
Remove theme_bw from visR

PR #181, opened 2021-05-21
Update readme

epijim:


epijim:
@timtreis - I added `vdiffr` to the suggests section so unit tests would run. Also downgraded `xslx` from depends to suggests as not sure it's needed for core functionality. 

I guess CRAN won't like how many dependencies there are (24?)

PR #187, opened 2021-05-23
clean up top level

bailliem:
clean up top level
up to date rbuild ignore


bailliem:
Delete logo.png file and removes docs folder that is not in use
add some files for pkgdown to build ignore.

PR #194, opened 2021-05-27
change behaviour of get risktable

SHAESEN2:
One issue is the alginment of the graph and the risktables. I thought I fixed it a few weeks ago but now the behaviour reverted again. We will need to fix this.

### DIFF
--- a/R/add_risktable.R
+++ b/R/add_risktable.R
@@ -49,6 +49,13 @@ add_risktable <- function(gg, ...){
 #'                       ,collapse = TRUE
 #'                       ,times = c(0,20,40,60)
 #'                      )
+#'                      
+#' ## Add risk set as specified times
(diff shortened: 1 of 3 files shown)


rebecca-albrecht (approved):

bailliem (commented):
Should we document this behavior to align ticks and times in plot and table as an example e.g.

```
km_fit <- survival::survfit(survival::Surv(AVAL, 1-CNSR) ~ TRTP, data=adtte)

visR::visr(km_fit, x_ticks = c(0, 20, 40, 100,111, 200)) %>% 
  add_risktable(times = c(0, 20, 40, 100,111, 200))
```

bailliem (approved):

rebecca-albrecht (approved):

PR #195, opened 2021-05-27
Bugfix/20210527 apply theme related

timtreis:
Not yet mergeable, just as a tool for easier resolving of merge conflicts. Will also be later used to address issues and to be merged back into develop.


R/visr_plot.R
```diff
@@ -92,10 +92,10 @@ visr.survfit <- function(
  ,y_ticks = NULL
  ,fun = "surv"
  ,legend_position = "right"
+ ,theme = "visR"
```
SHAESEN2, line 95: Our current thinking was to set a global theme instead of using a theme argument for every plot. We also agreed no more features as of last week.

R/visr_plot.R
```diff
@@ -258,10 +259,24 @@ visr.survfit <- function(
                                 limits = c(min(y_ticks), max(y_ticks))) +
     ggplot2::theme(legend.position = legend_position) +
     NULL
-
+  
   class(gg) <- append(class(gg), "ggsurvfit")
   
-  return(gg)
+  if (theme == "visR") {
```
SHAESEN2, line 265: after cran:

look for a globally defined theme => if not present go for visR theme

tests/testthat/test-apply_theme.R
```diff
@@ -1,12 +1,12 @@
...
+#' @title Specifications apply_theme
 #' @section Last updated by:
 #' Tim Treis
 #' @section Last update date:
 #' 07-MAY-2021
 
 # Specifications ---------------------------------------------------------------
 
-#' T1. The `define_theme` function returns an object that has a valid shape for usage in the `add_theme` function.
+#' T1. The `define_theme` function returns an object that has a valid shape for usage in the `apply_theme` function.
```
SHAESEN2, line 9: What is a valid shape? Can you be more specific?

tests/testthat/test-apply_theme.R
```diff
@@ -1,12 +1,12 @@
...
 #' T1.1 No error when no parameters are specified.
 #' T1.2 Not specifying any parameters returns a list.
```
SHAESEN2, line 11: an empty list?

tests/testthat/test-apply_theme.R
```diff
@@ -1,12 +1,12 @@
...
 #' T1.3 No error when `strata` is `NULL`.
```
SHAESEN2, line 12: Im confused why a theme needs to know strata.

tests/testthat/test-apply_theme.R
```diff
@@ -156,15 +156,15 @@ testthat::test_that("T1.18 A warning when `bg` is anything but a character.",{
...
+testthat::context("apply_theme - T2. The `apply_theme` function applies the specified changes to a `ggplot` object.")
 
 testthat::test_that("T2.1 No error when a `ggplot` plot is provided, but no theme.",{
   
   gg <- adtte %>%
     visR::estimate_KM("SEX") %>%
     visR::visr()
   
-  testthat::expect_error(visR::add_theme(gg), NA)
+  testthat::expect_error(visR::apply_theme(gg), NA)
```
SHAESEN2, line 167: this does not prove that the changes have been applied, only that there are no errors generated

tests/testthat/test-apply_theme.R
```diff
@@ -25,15 +25,15 @@
...
 #' T2.2 No error when a `ggplot` plot and a minimal `visR::define_theme` object are provided.
 #' T2.3 No error when a `ggplot` plot and a complex `visR::define_theme` object are provided.
 #' T2.4 A message when a theme not generated through `visR::define_theme` is provided.
 
 # Requirement T1 ---------------------------------------------------------------
 
-context("add_theme - T1. The `define_theme` function returns an object that has a valid shape for usage in the `add_theme` function.")
+testthat::context("apply_theme - T1. The `define_theme` function returns an object that has a valid shape for usage in the `apply_theme` function.")
 
 testthat::test_that("T1.1 No error when no parameters are specified.",{
```
SHAESEN2, line 39: Here I still dont see what a 'valid' shape is

tests/testthat/test-apply_theme.R
SHAESEN2: you will need to go through ggbuild and see if the theme elements are there

R/apply_theme.R
```diff
@@ -9,6 +9,7 @@
 #' @param fontfamily string with the name of a supported font
 #' @param grid boolean that specifies whether the grid should be drawn or not
 #' @param bg string giving the colour for the background of the plot
+#' @param legend_position string indicating the legend position
```
SHAESEN2, line 12: We decided to remove @author everywhere. I know we have not been consistent in this :(

R/apply_theme.R
```diff
@@ -9,6 +9,7 @@
 #' @param fontfamily string with the name of a supported font
```
SHAESEN2, line 9: fontsizes: list for sizes for different options is a bit obscure. I would not know how many font sizes I need to provide based on this, nor to what it will be applied. Can you be a bit more specific?

SHAESEN2: The same is true for strata. You will need to provide a better explanation

R/apply_theme.R
```diff
@@ -20,18 +21,23 @@
...
 #'
 #' @export
 
 define_theme <- function(strata = NULL,
                          fontsizes = NULL,
                          fontfamily = "Helvetica",
                          grid = FALSE,
-                         bg = "transparent") {
+                         bg = "transparent",
+                         legend_position = NULL) {
```
SHAESEN2, line 40: Putting strata to NULL is not our default in visR. If no strata are provided, we have 'Overall'. Is strata argument required? If not filled out we can default to ggplot2 colors right? I dont see the need for a warning.

SHAESEN2: The same is true for font sizes etc ... If the user does not provide this, we can provide a default or take the one of ggplot2

R/apply_theme.R
```diff
@@ -90,14 +96,36 @@ define_theme <- function(strata = NULL,
...
+    
+  } else if (is.list(grid)) {
+    
+    if (("major" %in% names(grid)) | ("minor" %in% names(grid))) {
+      
+      theme[["grid"]] <- grid
+      
+    } else {
+      
+      base::warning("Invalid argument for `grid`. Please use a boolean or a list to indicate whether you want a background grid. Setting to default `FALSE`.")
```
SHAESEN2, line 120: Also here. I'm not a fan of giving warnings for every argument This is really annoying as a user because you did not do anything wrong.

SHAESEN2: Default behaviour should be described in roxygen2 documentation

SHAESEN2 (changes requested):
Several obscure descriptions and testing present.

No more features allowed in existing functions untill after CRAN release.

Please include spacing of labels with respect to axis as you removed "\n" in visr plotting.

R/apply_theme.R
timtreis: Well, that's why the warnings are only triggered if the user did anything wrong :D If you don't manually specify it, it's `FALSE` and doesn't trigger a warning. If you manually set it to `TRUE` or `FALSE` you don't get a warning. If you set it to a list that has at least one of ["major", "minor"] as members, you don't get a warning. Only if none of these cases are true, a warning is given.

R/apply_theme.R
timtreis: Which is exactly what is happening when you call it without any parameters. Them being set to NULL was supposed to be more of a way to expose the options that will be evaluated.
image

R/apply_theme.R
timtreis: Will do :+1:

R/apply_theme.R
timtreis: Oh, okay. Wasn't present in that meeting. Should I also remove them for the other files in this PR?

image

tests/testthat/test-apply_theme.R
timtreis: Will improve the wording :+1:

tests/testthat/test-apply_theme.R
timtreis: Will improve the wording :+1:

tests/testthat/test-apply_theme.R
timtreis: True :thinking: Do we already use this non-vdiffr strategy in some other function?

R/apply_theme.R
SHAESEN2: I prefer errors if a user misspecifies an argument but we can discuss after CRAN release.

tests/testthat/test-apply_theme.R
SHAESEN2: I think so but I cant remember. You can do a look in files for ggbuild

R/apply_theme.R
SHAESEN2: Yes that is good or ask Mark to do so to avoid merge conflicts. Thanks for that

tests/testthat/test-apply_theme.R
timtreis: image
No, some defaults are (silently) automatically filled in.

R/apply_theme.R
cschaerfe: Aactually, in `estimate_KM` and `get_tableone` we are setting the `strata` parameter to NULL as default which is treated internally as "overall" in the sense that data is not stratified

R/apply_theme.R
cschaerfe: Good discussions. Let's agree on a common theme. I agree that that there should be _some_ action triggered when user does something wrong and alternative behaviour is applied. Otherwise it's a nightmare to debug. 
The question is if misspecifying parameters should actually cause the function to stop or merely warn. ggplot2 often just warns, depending on the severity of the issue.

tests/testthat/test-apply_theme.R
```diff
@@ -0,0 +1,407 @@
...
+  
+})
+
+testthat::test_that("T1.13 No error when `fontfamily` is a string.",{
+  
+  testthat::expect_error(visR::define_theme(fontfamily = "Times"), NA)
+  
+})
+
+testthat::test_that("T1.14 A warning when `fontfamily` is anything but a string.",{
```
cschaerfe, line 126: what if fontfamily is a list of strings?

tests/testthat/test-apply_theme.R
```diff
@@ -0,0 +1,407 @@
...
+testthat::test_that("T1.16 A warning when `grid` is anything but a boolean",{
+  
+  testthat::expect_warning(visR::define_theme(grid = NULL))
+  testthat::expect_warning(visR::define_theme(grid = 12))
+  testthat::expect_warning(visR::define_theme(grid = "visR"))
+  testthat::expect_warning(visR::define_theme(grid = c()))
+  
+})
+
+testthat::test_that("T1.17 No error when `bg` is a character.",{
```
cschaerfe, line 151: any kind of character? or a valid color string?

cschaerfe (approved):

tests/testthat/test-apply_theme.R
timtreis: Good catch, fixed it

tests/testthat/test-apply_theme.R
timtreis: Right now, any kind of character. We pass this through to ggplot and let them figure out what a valid colour is.

PR #199, opened 2021-05-31
Test CI/CD develop to main checks

bailliem:
Test to check CICD runs merging to master. 

Want to check that we can merge to master from a clean develop and that documentation automatically builds.


DESCRIPTION
```diff
@@ -29,11 +30,11 @@ Suggests:
     vdiffr,
     knitr,
     rmarkdown,
-    survminer,
     forcats,
     glue,
-    here,
-    stringr
+    stringr,
```
SHAESEN2, line 35: if we can avoid this one, please do. It does not have added value as all string manipulations are straightforward with base R

DESCRIPTION
```diff
@@ -29,11 +30,11 @@ Suggests:
...
+    openxlsx,
+    data.table
```
SHAESEN2, line 37: we dont need this. If you get an error because of this, we need to solve it.

DESCRIPTION
```diff
@@ -45,14 +46,10 @@ Imports:
...
     tibble,
     tidyr,
     dplyr,
-    ggpubr,
-    pammtools,
     kableExtra,
-    assertthat,
     gt (>= 0.1.0),
     tidyselect,
     parcats (>= 0.0.1),
```
SHAESEN2, line 55: we dont need this one. as.factor and factor should be sufficient from base R

DESCRIPTION
```diff
@@ -62,7 +59,5 @@ Imports:
     gridExtra,
     plotly,
```
SHAESEN2, line 60: not yet required?

R/add_risktable.R
```diff
@@ -31,35 +31,32 @@ add_risktable <- function(gg, ...){
 #' 
 #' @examples
 #'
+#' library(survival)
```
SHAESEN2, line 34: remove

SHAESEN2 (changes requested):

DESCRIPTION
bailliem: Removed both from the release branch DESCRIPTION file

DESCRIPTION
bailliem: Removed

DESCRIPTION
bailliem: Removed pltly. gridExtra is used in add_annotation

PR #200, opened 2021-05-31
Fixed issue in which the data source got duplicated when interacting.

timtreis:
In response to an (ancient) open issue.


timtreis:
I preemptively deleted an issue and then @cschaerfe and I found out that it was a bug with a quick fix. Is this still allowed before release?

cschaerfe (approved):

PR #204, opened 2021-06-01
Don't merge yet, place for discussion

timtreis:


vignettes/Styling_KM_plots.Rmd
```diff
@@ -0,0 +1,136 @@
...
+
+## Using `visR` to style the plot
+
+The most direct option to style plots generated through `visR::visr()` is by using the parameters that the function provides. Internally, parameters like the y-axis label are automatically deducted from the used function. The following example demonstrates the options exposed.
+
+```{r visr-parameter-styling}
+lung_suvival_object %>%
+  visR::visr(x_label = "Time",
+             y_label = NULL, # NULL (default) causes the label to be deducted from the used function
+             x_units = "days",
```
SHAESEN2, line 80: x axis does not contain " days"

vignettes/Styling_KM_plots.Rmd
```diff
@@ -0,0 +1,136 @@
...
+             x_ticks = seq(0, 1200, 200),
+             y_ticks = seq(0, 100, 20),
+             fun = "pct",
+             legend_position = "top") 
+```
+However, these rather minimal adjustments usually don't cover all the things a user wants to modify. Therefore, we provide two additional function to adjust more aesthetics: `visR::define_theme()` and `visR::apply_theme()`. The first one provides an easy wrapper to create a nested list of list with styling options that is then applied to the plot by the second function.
```
SHAESEN2, line 86: two additional function**s**

SHAESEN2: = typo

vignettes/Styling_KM_plots.Rmd
```diff
@@ -0,0 +1,136 @@
...
+
+### Defining a `visR_theme` using `visR::define_theme()`
+If no further options are provided to `visR::define_theme()`, it nonetheless returns a very minimal list of reasonable defaults.
+```{r visr-define_theme-empty}
+visR::define_theme() 
+```
+However, this function also takes several other styling options. The currently usable ones are displayed below. One particular use that we had in mind when writing this function was, that we wanted to have the option to define the different colours for the strata once and then to not have to worry about all of them being present. When defining the strata through a list of colours, the absence of data for a particular option would shift the order.
```
SHAESEN2, line 93: If the strata are factors as they are expected, then no shift would happen. I would rephrase this a bit.

SHAESEN2 (changes requested):

bailliem (approved):

PR #211, opened 2021-06-05
Bugfix/20210605 xlabel

timtreis:


R/visr.R
```diff
@@ -235,11 +235,18 @@ visr.survfit <- function(
 
 # Obtain X-asis label ----------------------------------------------------------
 
-  if (is.null(x_label)){
+  if (is.null(x_label)) {
+    
     if ("PARAM" %in% names(x)) x_label = x[["PARAM"]]
     if (! "PARAM" %in% names(x)) x_label = "time"
```
SHAESEN2, line 241: Then x_label = what the user specifies. Don't force "time".

R/visr.R
```diff
@@ -235,11 +235,18 @@ visr.survfit <- function(
...
     if (!is.null(x_units)) x_label = paste0(x_label, " (", x_units, ")")
```
SHAESEN2, line 242: this does not make sense. if the label is not specified, then units should not be added as well. You agree looking at the testing script :)

tests/testthat/test-visr_plot.R
```diff
@@ -332,6 +333,27 @@ testthat::test_that("T2.15 An error when an undefined option is passed to `legen
   
 })
 
+testthat::test_that("T2.16 When both `x_label` and `x_units` are specified, they are concatenated into the x-axis label.", {
```
SHAESEN2, line 336: add tests for one of them not specified and a test for both not specified

SHAESEN2 (changes requested):

R/visr.R
timtreis: Well, I'm not :D 

If `x_label` and `x_units` are both not `NULL`, they are concatenated to `<x_label> (<x_units>)`.

If `x_label` is not `NULL`, but `x_units` is it set's the label to `x_label`.

If `x_label` is `NULL`, but `x_units` is not, it tries to deduct the name from the data. If that is not possible, it's set to the default "time". And then it appends `x_units`.

If `x_label` and `x_units` are `NULL`, it tries to deduct the name from the data. If that is not possible, it's set to the default "time".

R/visr.R
timtreis: I think it does make sense, as for example in the example below.

```
adtte %>%
  visR::estimate_KM("SEX") %>%
  visR::visr(x_label = NULL,
             x_units = "days")
```
image

tests/testthat/test-visr_plot.R
timtreis: T2.1 and T2.3 cover this behaviour since their default is `NULL`.

R/visr.R
SHAESEN2: Dont use "time" as default. Default is NULL.

R/visr.R
SHAESEN2: Sure but here the label is extracted from the dataset. When no PARAM is available then it does not make sense.

timtreis:
Changes that were to introduced here were now part of [PR 212](https://github.com/openpharma/visR/pull/212) which makes this PR unnecessary.

Issue #207, opened 2021-06-02
Not all ggplot2 theming options are actually applied to our plot


timtreis:
#### Issue description

Not all ggplot2 aesthetics modifications are applied to the plot. 

#### Steps to reproduce the issue

```r
library(visR)
library(survival)
library(ggplot2)

lung_cohort <- survival::lung

lung_cohort <- lung_cohort %>%  
  dplyr::mutate(sex = as.factor(ifelse(sex == 1, "Male", "Female")))  %>%  
  dplyr::mutate(status = status - 1) %>%
  dplyr::rename(Age = "age", Sex = "sex", Status = "status", Days = "time") 

lung_suvival_object <- lung_cohort %>%
  visR::estimate_KM(strata = "Sex", CNSR = "Status", AVAL = "Days")

p <- lung_suvival_object %>%
  visR::visr()

p +
  ggplot2::theme_bw() +
  ggplot2::xlab("Time (days)")
```

#### What's the expected result?

- The x-axis label is changed to "Time (days)"

#### What's the actual result?

- The x-axis label is unchanged.

PR #212, opened 2021-06-05
Changed the way the x- and y-axis labels are applied

timtreis:
Very minor change that however fixes the previous incompatibility with ggplot2s' xlab/ylab commands.

Adresses issue https://github.com/openpharma/visR/issues/207

Rough troubleshooting train of thought:
a) By inspecting the plot components I noticed that the desired label actually ends up in the plot data.
b) Experiment around with priorities of ggplot scales and labels
c) We're currently setting the labels through the scales which seems to make them "write-protected" to the xlab/ylab commands.

### DIFF
--- a/R/visr.R
+++ b/R/visr.R
@@ -235,11 +235,42 @@ visr.survfit <- function(
 
 # Obtain X-asis label ----------------------------------------------------------
 
-  if (is.null(x_label)){
-    if ("PARAM" %in% names(x)) x_label = x[["PARAM"]]
-    if (! "PARAM" %in% names(x)) x_label = "time"
-    if (!is.null(x_units)) x_label = paste0(x_label, " (", x_units, ")")
+  if (base::is.null(x_label)){
+    
+    if ("PARAM" %in% base::names(x)) {
+      
+      if (base::length(base::unique(x[["PARAM"]])) == 1) { 
+        
+        x_label <- as.character(x[["PARAM"]][[1]])
+        
+      } else {
+        
+        base::warning("More than one unique entry in 'PARAM'.")
+        
+      }
+      
+    } else if ("PARAMCD" %in% base::names(x)) {
+      
+      if (base::length(base::unique(x[["PARAMCD"]])) == 1) { 
+        
+        x_label <- as.character(x[["PARAMCD"]][[1]])
+        
+      } else {
+        
+        base::warning("More than one unique entry in 'PARAMCD'.")
+        
+      }
+      
+    } else {
+      
+      base::warning("The x-axis label was not specified and could also not be automatically determined due to absence of 'PARAM' and 'PARAMCD'.")
+      
+    }
+    
+    if (!is.null(x_units)) { x_label = paste0(x_label, " (", x_units, ")") }
+    
   }
+  
   if (is.null(x_ticks)) x_ticks = pretty(x$time, 10)
 
 # Obtain Y-asis label ----------------------------------------------------------
--- a/R/visr.R
+++ b/R/visr.R
@@ -268,13 +268,13 @@ visr.survfit <- function(
 
   gg <- ggplot2::ggplot(tidy_object, ggplot2::aes(x = time, group = strata)) +
     ggplot2::geom_step(ggplot2::aes(y = est, col = strata)) + 
-    ggplot2::scale_x_continuous(name = x_label,
-                                breaks = x_ticks,
+    ggplot2::scale_x_continuous(breaks = x_ticks,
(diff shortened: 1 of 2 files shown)


tests/testthat/test-visr_plot.R
```diff
@@ -199,11 +199,13 @@ testthat::test_that("T2.8 When `x_label` is `NULL` and the `survfit` object does
...
   
-  gg <- survfit_object %>% visR::visr(y_label = NULL)
+  gg <- survfit_object %>% visR::visr(x_label = NULL)
   
   ggb <- ggplot2::ggplot_build(gg)
   
-  testthat::expect_true("time" %in% ggb$layout$panel_params[[1]]$x$name)
+  # Ensures that the label is pulled from gg$labels$x
+  testthat::expect_true("waiver" %in% class(ggb$layout$panel_params[[1]]$x$name))
+  testthat::expect_true("time" %in% gg$labels$x)
```
SHAESEN2, line 208: Why do we need time in the label. The user is allowed to choose, we cant force that.

R/visr.R
```diff
@@ -268,13 +268,13 @@ visr.survfit <- function(
 
   gg <- ggplot2::ggplot(tidy_object, ggplot2::aes(x = time, group = strata)) +
     ggplot2::geom_step(ggplot2::aes(y = est, col = strata)) + 
-    ggplot2::scale_x_continuous(name = x_label,
-                                breaks = x_ticks,
+    ggplot2::scale_x_continuous(breaks = x_ticks,
                                 limits = c(min(x_ticks), max(x_ticks))) +
-    ggplot2::scale_y_continuous(name = y_label,
-                                breaks = y_ticks,
+    ggplot2::xlab(x_label) +
```
SHAESEN2, line 273: is this for the theming? Would be helpful to put a rationale in your commit :)

tests/testthat/test-visr_plot.R
```diff
@@ -199,11 +199,13 @@ testthat::test_that("T2.8 When `x_label` is `NULL` and the `survfit` object does
   
   names(survfit_object)[which(names(survfit_object) == "PARAM")] <- "visR"
   
-  gg <- survfit_object %>% visR::visr(y_label = NULL)
+  gg <- survfit_object %>% visR::visr(x_label = NULL)
```
SHAESEN2, line 202: good catch

SHAESEN2 (changes requested):

R/visr.R
timtreis: Oh sorry, I should've copied the problem from the linked issue :D

tests/testthat/test-visr_plot.R
timtreis: image

Because this "time" is the default given name and the test checks for that

SHAESEN2: I dont agree with a default name. If the user does not specify anything and param in not in the dataset, then default should be NULL

SHAESEN2 (changes requested):

SHAESEN2:
Hey Tim, Did you also include additional testing for xlabel and xunits logic?

tests/testthat/test-visr_plot.R
timtreis: So leaving the axis unlabeled? I feel like this goes against our biased approach, no? Maybe we can talk about that during the next core group meeting :+1:

timtreis:
> Hey Tim, Did you also include additional testing for xlabel and xunits logic?

image


Yeah :+1: The use cases in which not both are specified are covered by other tests imho

tests/testthat/test-visr_plot.R
timtreis: I created an issue for it so we don't forget: https://github.com/openpharma/visR/issues/213

SHAESEN2:
We also need to check if there is only param/paramcd present in the data before plotting

timtreis:
> We also need to check if there is only param/paramcd present in the data before plotting

Oh true. Right now I'm only checking for it getting overwritten, but not for it being deducted from the data. I'll also write that test.

tests/testthat/test-visr_plot.R
SHAESEN2: Thanks. Leaving the axis unlabeled is better because you will see you missed something. We can trigger a warning.

bailliem:
Add a warning that x/y label is not populated. Warning to console. with proposal to populate with empty string if that is the intention

SHAESEN2 (approved):
Nice work!

Issue #222, opened 2021-06-10
Ensure all function documentation describes return value


bailliem:
#### Issue description

Please add \value to .Rd files regarding exported methods and explain
the functions results in the documentation. Please write about the
structure of the output (class) and also what the output means. (If a
function does not return a value, please document that too, e.g.
\value{No return value, called for side effects} or similar)
Missing Rd-tags in up to 18 .Rd files, e.g.:
      get_risktable.Rd: \value
      get_tableone.Rd: \value
      pipe.Rd: \value
      render.Rd: \value
      summarize_long.default.Rd: \value
      summarize_long.factor.Rd: \value

PR #224, opened 2021-06-10
Issue/222 value docs

bailliem:
To address issue #222


rebecca-albrecht (approved):

timtreis (changes requested):
\value missing for some of the functions. I'm in the process of adding these.

timtreis (approved):
Now all functions should have a @return -> \value tag.

Also fixed a weird formatting error in `table_caption`

Issue #234, opened 2021-07-02
'riskmetric' doesn't recognise our vignettes


timtreis:
#### Issue description

When assessing `visR` with `riskmetric`, it doesn't recognise the existence of our vignettes.

#### Steps to reproduce the issue

![image](https://user-images.githubusercontent.com/12818242/124235931-2a472080-db16-11eb-81db-f9fddcc03956.png)

#### What's the expected result?

- We do have vignettes, so it should be 1

#### What's the actual result?

- It's 0

timtreis:
1) `riskmetric::pkg_ref("visR")` creates a `pkg_ref` object and populates it with the scores to be determiend.

timtreis:
Reinstall solved it.

PR #235, opened 2021-07-02
Add riskmetric badge to readme

### LINKED ISSUE #232: Add in the packages score on a risk metric
## Idea

Could we track the 'trust worthiness' of the package as a badge? e.g.  calculate risk metric score and show it as a badge? 

## Steps

Add riskmetric to cicd - and expose as badge (if no badge service exists I guess could just write a svg via the gh action).

## Concern

Is the score designed to be used on a single package? e.g. what does '0.8' mean to an end user?

It's also easy for us to game the number by looking at the categories - although maybe that also increases the quality of the package so is ok?

epijim:
FYI for #234 and closes #232

### DIFF
--- a/.github/workflows/makedocs.yml
+++ b/.github/workflows/makedocs.yml
@@ -33,6 +33,7 @@ jobs:
           install.packages("vdiffr")
           install.packages("textshaping", type = "binary")
           remotes::install_github("lorenzwalthert/gitsum")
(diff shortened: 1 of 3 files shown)


timtreis:
How do you calculate the score? Locally I get a score of 0.37. The entire riskmetric calculation doesn't seem to be machine-independent. I already raised an issue for this -> https://github.com/pharmaR/riskmetric/issues/206

PR #237, opened 2021-07-06
testthat file dedicated to verify compliance with CRAN guidelines 

timtreis:
This test automatically scans the relevant parts of our codebase for possible violations of additional style guidelines we received during the CRAN submission process. Generally, a list of files is retrieved which is then searched analysed using a regex pattern specific to the respective issue. If found, the test fails and the file and line of the issue are printed into the test log.

```
#' T1. Our codebase doesn't violate CRAN style-guidelines.
#' T1.1 TRUE/FALSE are used instead of T/F.
#' T1.2 Each function documentation contains a \value{} tag.
#' T1.3 The existence of packages is not checked through 'installed.packages()'.
#' T1.4 No \dontrun{} tags unless the code actually takes a long time.
#' T1.5 The use of 'options()' is immediately preemptively reverted.
```

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -43,6 +43,79 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
   
 }
 
+#' A helper function to get all files in the ~/R/ folder
+#' @keywords internal
+
+get_files_in_R_folder <- function() {
+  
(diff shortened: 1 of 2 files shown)


bailliem:
Nice branch name

bailliem (approved):

PR #250, opened 2021-08-02
Feature: add_highlight()

timtreis:


R/add_CI.R
```diff
@@ -105,12 +107,12 @@ add_CI.ggsurvfit <- function(gg,
...
-                                        fill = "transparent", 
-                                        colour = strata),
-                           alpha = alpha, 
+                                        fill = strata, 
+                                        colour = strata,
+                                        alpha = alpha),
                            outline.type = "both",
                            linetype = linetype, 
-                           show.legend = FALSE,
+                           show.legend = TRUE,
```
SHAESEN2, line 115: should be FALSE. Nobody wants to see color around the linetype in a clinical setting. We can discuss and make this an optional argument, but TRUE should not be the default.

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
+#' @title Highlight a specific strata
+#'
+#' @description S3 method for highlighting a specific strata by lowering the opacity of all other strata.
+#'
+#' @author Tim Treis
```
SHAESEN2, line 5: we agreed to remove author

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#'
+#' @param gg visR object
+#' @param ... other arguments passed on to the method 
+#' 
+#' @examples
+#' library(visR)
```
SHAESEN2, line 11: examples should not have library calls

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#'
+#' adtte %>%
+#'   visR::estimate_KM("SEX") %>%
```
SHAESEN2, line 14: strata = "SEX" (examples should be clear)

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#'   visR::visr() %>%
+#'   visR::add_CI(alpha = 0.4) %>%
+#'   visR::add_highlight("SEX=M", bg_alpha_multiplier = 0.2)
```
SHAESEN2, line 17: bg_alpha_mulitplier: 
- too long argument name
- you said you were going to decrease opacity of other strata so mulitplier is confusing
- perhaps we can do something like "alpha = xx" for the strata you want to highlight and you reduce accordingly

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#'
+#' @return The input `ggplot2` object with adjusted `alpha` values
```
SHAESEN2, line 19: input will be of ggsurvfit not purely ggplot2 :)

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#' @rdname add_highlight
+#'
+#' @export
+
+add_highlight <- function(gg, ...){
+  UseMethod("add_highlight", gg)
+}
+
+#' @param gg A ggplot created with visR
+#' @param strata Name of the strata to be highlighted as shown in the legend.
```
SHAESEN2, line 30: String representing the name and value of the strata ...

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+#' @param bg_alpha_multiplier Factor with which the `alpha` values of all but the specified strata will be multiplied.
```
SHAESEN2, line 31: you wanted to decrease opacity so this is confusing. Can you rephrase?

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+add_highlight.ggsurvfit <- function(gg, 
+                                    strata, 
+                                    bg_alpha_multiplier = 0.2,
+                                    ...) {
+  
+  # Ugly hack to suppress CRAN warning as described here:
+  # https://www.r-bloggers.com/2019/08/no-visible-binding-for-global-variable/#option-two
+  alpha <- colour <- fill <- group <- NULL
+  
+  if (missing(strata)) {
```
SHAESEN2, line 46: Can you also check for the structure of strata? It needs to contain an actual strata, part of the data. It needs to contain "=" and it needs to contain one of the values of the strata.
You could remove all blanks within the strata argument before working with the argument to build in some defensiveness eg sex =  M will not be recognized => make it sex=M at the start.

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+  gg_gb <- ggplot2::ggplot_build(gg)
+  gg_gtable <- ggplot2::ggplot_gtable(gg_gb)
+  gg_guidebox_id <- base::which(sapply(gg_gtable$grobs, function(x) x$name) == "guide-box")
+  gg_table_grob <- gg_gtable$grobs[[gg_guidebox_id]]$grobs[[1]]
+  
+  # Get IDs of elements containing strata labels
+  strata_label_ids <- base::grep("label", gg_table_grob$layout$name)
+  strata_labels <- c()
+  
+  for (id in strata_label_ids) {
```
SHAESEN2, line 63: use functional programming instead of loops

R/add_highlight.R
```diff
@@ -0,0 +1,136 @@
...
+    
+    warning(msg)
+    return(gg)
+    
+  } else {
+    
+    # Which group(s) in the ggplot data object corresponds to the bg strata?
+    bg_strata_ids <- base::which(strata != strata_labels)
+    
+    # Replace the previous hex alpha values with the new one
```
SHAESEN2, line 87: use functional programming instead of loops

R/visr.R
```diff
@@ -301,7 +301,9 @@ visr.survfit <- function(
 
   yscaleFUN <- function(x) sprintf("%.2f", x)
 
-  gg <- ggplot2::ggplot(tidy_object, ggplot2::aes(x = time, group = strata)) +
+  gg <- ggplot2::ggplot(tidy_object, ggplot2::aes(x = time,
```
SHAESEN2, line 304: why add the fill = strata argument?

R/utils_visr.R
```diff
@@ -267,3 +267,143 @@ legendopts <- function(legend_position = "right",
 
 }
 
+#' @title Extract the numerical alpha representation of #RRGGBBAA colour
```
SHAESEN2, line 270: why do we need below functions? They seem very similar. Could we not make one big function? Perhaps I misunderstood?

SHAESEN2 (changes requested):

R/add_highlight.R
timtreis: Right now I am expecting the user to use exactly what is shown in the legend and present him with the exact options he can pick. I think that is more robust than trying to guess what the user meant, no?
image

R/add_highlight.R
timtreis: I think this is a bit of a tricky issue.

I think the route of specifying the foreground stratas' `alpha` and then automatically reducing the others would be too strong of an opinion because the user might want to adjust this, depending on preference and colourscheme.

While I am decreasing the `alpha` of the background strata, I have written the code with `add_CI` in mind. Since the CI already have an `alpha`, chose to go with a multiplier instead of subtraction. This way, both the line as well as the CI are made more transparent proportionally.

But I fully agree that the name is too long :D I just wanted the name to be self-explanatory and didn't come up with something better yet.

R/add_CI.R
timtreis: Both, `TRUE` and `FALSE` actually don't have an influence on the outcome here so I'll remove it.

R/utils_visr.R
timtreis: ggplot seems to use a mixture of a column called "alpha" and appending the alpha value to a #RRGGBB colour in the "colour" column to adjust the alpha values, depending on the user behaviour. Therefore I needed helper functions to convert between numerical and hex-encoded alpha values both ways (`convert_alpha()`), extract the numerical alpha value of a #RRGGBBAA colour to apply my modifier to it (`get_alpha_from_hex_colour()`) and a convencience wrapper to replace the AA part of a #RRGGBBAA colour (`replace_hex_alpha()`). Some of the functions use each other, but cramming all the functionality into one big function would make it very cumbersome to use.

R/visr.R
timtreis: If this is not specified at a `ggplot2::ggplot()` level, I couldn't find any other way to adjust the alpha values after plotting. Now I can extract the current colours, including their alphas, and apply the modified values as a `scale_colour_manual()`/`scale_fill_manual()`. This way, I only modify the actual ggplot object without creating a new one from scratch afterwards.

R/add_highlight.R
timtreis: I think this is the most accurate description of what it does. Yes, it allows only values between 0 and 1 as input and can therefore never increase the alpha, but technically it's just a multiplication.

See also the answer to your other comment:

> 
> I think the route of specifying the foreground stratas' alpha and then automatically reducing the others would be too strong of an opinion because the user might want to adjust this, depending on preference and colourscheme.
> 
> While I am decreasing the alpha of the background strata, I have written the code with add_CI in mind. Since the CI already have an alpha, I chose to go with a multiplier instead of subtraction. This way, both the line as well as the CI are made more transparent proportionally.

R/utils_visr.R
SHAESEN2: I would explain this behaviour in a bit more detail in the description so we remember this and users understand it as well.

R/add_CI.R
```diff
@@ -105,12 +107,11 @@ add_CI.ggsurvfit <- function(gg,
...
-                                        colour = strata),
-                           alpha = alpha, 
+                                        fill = strata, 
+                                        colour = strata,
+                                        alpha = alpha),
                            outline.type = "both",
                            linetype = linetype, 
-                           show.legend = FALSE,
                            na.rm = TRUE)
   }
```
SHAESEN2, line 117: Can you put back the default where the CI does not color the legend surrounding the step function? This is not useful for clinical reporting. When no CI is requested we see a grey background surrounding the step function in the legend. Can this be white? It will be difficult to see when printed out.

R/add_highlight.R
SHAESEN2: I still don't agree:
- make argument name shorter
- make it more meaningful as to what is happening

SHAESEN2 (changes requested):

R/add_CI.R
timtreis: I removed the grey background in the legend in `visr.R`. Do we also want to get rid of the coloured background that we get when adding CIs?

SHAESEN2: Yes I think that makes sense.

SHAESEN2 (changes requested):

R/add_CI.R
timtreis: Now no bg should ever be displayed.

Issue #258, opened 2021-09-12
get_files_in_xxx_folder - remove


SHAESEN2:
We should consider removing these are they are not true functions but calls to predefined codes.
We can decide to keep them if they are used many times across multiple files.
If we decide to keep them, could we shorten the name eg get_xx_files (get_rd_files, get_man_files, ...) in agreement with our naming conventions?

SHAESEN2:
Renaming was a good idea.

Creating functions that call upon one R command is not good practice. An alternative is to create a more general function eg get_files and one of the arguments is which folder you wish to extract files from.

PR #268, opened 2021-10-11
improved functions for gathering files to run CRAN watchdog on

timtreis:
As discussed in https://github.com/openpharma/visR/issues/258

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -59,7 +59,7 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
 
   if (functions) {
 
-    R_files <- list.files(path = paste0(wd, "/../../R"), 
+    R_files <- list.files(path = file.path(wd, "/../../R"), 
                           pattern = "*.R", 
                           full.names = TRUE)
     files <- c(files, unlist(R_files))
@@ -77,7 +77,7 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
 
   if (documentation) {
 
-    man_files <- list.files(path = paste0(wd, "/../../man"), 
+    man_files <- list.files(path = file.path(wd, "/../../man"), 
                             pattern = "*.Rd", 
                             full.names = TRUE)
     files <- c(files, unlist(man_files))
@@ -85,7 +85,7 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
 
   if (vignettes) {
 
-    vignette_files <- list.files(path = paste0(wd, "/../../vignettes"),
+    vignette_files <- list.files(path = file.path(wd, "/../../vignettes"),
                                  pattern = "*.Rmd", 
                                  full.names = TRUE)
     files <- c(files, unlist(vignette_files))


tests/testthat/helper.R
```diff
@@ -43,56 +43,16 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
-
-get_vignette_files <- function() {
-
-  vignette_files <- base::list.files(path = base::paste0(base::getwd(), "/../../vignettes"),
-                                     pattern = "*.Rmd",
-                                     full.names = TRUE)
-
-  return(unlist(vignette_files))
-
+get_files <- function(path, pattern) {
```
SHAESEN2, line 50: This is basically the same as you did before and not an improvement. I was thinking more in lines of:

get_files(path = base::file.path(wd, "/../../R"), pattern = "*.R") => return the vector instead of the list which is created by list.files.

However still not seeing any added value in a wrapper for one line of code.

SHAESEN2 (changes requested):

SHAESEN2:
Although I agree it reads better, it will be much harder to maintain the code and debug for beginners.
Survminer is also more readable but a pain to find what you need.

tests/testthat/helper.R
timtreis: I don't quite see what change you want here. I want to keep the `get_visR_files()` function since I am thinking of using it for another check on our package. But for the things going into it, I'm pretty neutral to it, just want to get it out of the way :D So do you want me to remove the (redundant) `unlist` or to just replace the `get_files` call directly with the `list.files` one?

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
-
-  return(unlist(test_files))
-
-}
-
-#' A helper function to get all files in the ~/man/ folder
-#' @keywords internal
-
-get_man_files <- function() {
+get_visR_files <- function(functions = FALSE,
```
SHAESEN2, line 49: For internal functions, I wonder if we should be consistent and use .get_visR_files

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
 
 }
 
-#' A helper function to get all files in the ~/R/ folder
+#' A helper function that returns the paths of the package files as a vector
```
SHAESEN2, line 46: The full path or relative paths? If relative, relative to what?

SHAESEN2 (changes requested):

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
-                                 full.names = TRUE)
-
-  return(unlist(test_files))
-
-}
-
-#' A helper function to get all files in the ~/man/ folder
-#' @keywords internal
+  files <- list()
+  wd <- getwd()
```
SHAESEN2, line 56: I dont see the purpose of this step

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
 
-get_man_files <- function() {
+  if (functions) {
 
-  man_files <- base::list.files(path = base::paste0(base::getwd(), "/../../man"),
-                                pattern = "*.Rd",
-                                full.names = TRUE)
+    R_files <- list.files(path = paste0(wd, "/../../R"),
```
SHAESEN2, line 60: use base::file.path

SHAESEN2: this does not work. 
> R_files
character(0)

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
+    R_files <- list.files(path = paste0(wd, "/../../R"), 
+                          pattern = "*.R", 
+                          full.names = TRUE)
+    files <- c(files, unlist(R_files))
 
-  return(unlist(man_files))
+  }
 
-}
+  if (tests) {
```
SHAESEN2, line 67: Dont you need the files from the test folder?

tests/testthat/helper.R
```diff
@@ -43,75 +43,54 @@ map_numbers_to_new_range <- function(numbers, lower, upper) {
...
 
-get_vignette_files <- function() {
-
-  vignette_files <- base::list.files(path = base::paste0(base::getwd(), "/../../vignettes"),
-                                     pattern = "*.Rmd",
-                                     full.names = TRUE)
+  }
 
-  return(unlist(vignette_files))
+  if (documentation) {
```
SHAESEN2, line 76: > .get_visR_files(documentation = T)
NULL

tests/testthat/helper.R
timtreis: Yes, because I think that if we already are enforcing the CRAN mandated conventions in some places, we should just enforce them in all places

PR #270, opened 2021-10-12
estimate_cox.R - wrapper for cox model

timtreis:
Hey Alexandra, I moved this file into a pull request until it is fully functional and tested :) Then we can merge it again.

I already fixed a few minor things, do you want to take over again?


R/estimate_cox.R
```diff
@@ -0,0 +1,195 @@
...
+#'
+#' ## Univariate Cox regression model using the covariate: `TRTP`
+#' # Stratified Kaplan-Meier analysis by `TRTP`
+#' visR::estimate_cox(data = adtte, main = "TRTP")
+#'
+#' ## Multivariate Cox regression model using the covariates: `TRTP` and `SEX`
+#' visR::estimate_cox(data = adtte, main = c("TRTP", "SEX"))
+#'
+#' ## Stratified Cox model (strata with one level)
+#' visR::estimate_cox(data = adtte, main = c("AGE", "survival::strata(PARAMCD)"))
```
SHAESEN2, line 36: i would add different argument for strata and for interactions

R/estimate_cox.R
```diff
@@ -0,0 +1,195 @@
...
+#'                CNSR = dplyr::if_else(status == 1, 0, 1)
+#'  )
+#'
+#' visR::estimate_cox(data = veteran_adam, main = c("trt", "diagtime"))
+
+
+
+estimate_cox <- function(
+  data = NULL
+  ,main = NULL
```
SHAESEN2, line 61: if you want to put all main, modifiers and interactions into one statment, then opt for consistency with estimate_KM and  call it strata

R/estimate_cox.R
```diff
@@ -0,0 +1,195 @@
...
+    stop(paste0("data can be of class `data.frame` or `tibble` or `data.table`."))
+  }
+
+  data <- as.data.frame(data)
+
+  # Validate columns --------------------------------------------------------
+
+  reqcols <- c(main, CNSR, AVAL)
+
+  # if (! all(reqcols %in% colnames(data))){
```
SHAESEN2, line 99: Why is this removed?

R/estimate_cox.R
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  if (!is.numeric(data[[CNSR]])) {
+    stop("Censor variable (CNSR) is not numeric.")
+  }
+
+
+  # Remove NA from the analysis --------------------------------------------
+
+  data <- as.data.frame(data) %>%
```
SHAESEN2, line 114: main should also be here

R/estimate_cox.R
```diff
@@ -0,0 +1,195 @@
...
+    survfit_object[["PARAMCD"]] <- paste(unique(data[["PARAMCD"]]), collapse = ", ")
+  }
+
+
+  # Return ------------------------------------------------------------------
+
+
+  summary_cox <- summary(survfit_object, conf.int = conf.int)
+
+ # res1 <- c(
```
SHAESEN2, line 175: The next section could go in a tidyme method

AlexandraP-21:
@timtreis thanks for noticing, this is because PARAMCD has only 1 category. I will change it to RACE which has more.

R/estimate_cox.R
AlexandraP-21: We cannot call it strata because strata we can have stratified parameters which will be denoted with 'strata' in the model. So we cannot have strata for the stratified parameters and strata for the model as well. But other ideas instead of main are more than welcome! :)

R/estimate_cox.R
AlexandraP-21: Because 'main' might have 'strata(RACE)' so that means that the if statement will start looking for 'strata(RACE)' and not just 'RACE'. I was trying to find a way to just get the parameters but I could not figure something out. Again any ideas more than welcome!

R/estimate_cox.R
SHAESEN2: We can have both. eg model(data=data) Can you give an example of why it would not work?

R/estimate_cox.R
SHAESEN2: Can you give a better example?

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
+#' @title Wrapper for Cox Proportional Hazard analysis
+#'
+#' @description This function is a wrapper around \code{survival::coxph} to fit a Cox Proportional Hazards Regression model, assuming right-censored data.
+#'    The function expects that the data has been filtered on the parameter (PARAM/PARAMCD) of interest.
+#'    The result is an object of class \code{summary.coxph} which can be used in downstream functions and methods that rely on the \code{summary.coxph} class.
+#'
```
SHAESEN2, line 6: #'    By default:
#'    \itemize{
#'      \item{A two-sided pointwise 0.95 confidence interval is estimated using a log transformation (conf.type = "log").}
#'      \item{The Efron approximation is used for tie handling.}
#'    }

@AlexandraP-21 Should we enforce some other SAS defaults? What are some of the other differences with SAS? Should we document them in this description as well? I believe it might be important for Pharma to understand those without investigating it themselves as we did. We have a presentation about those differences ;)

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+#' @param data The name of the dataset for the Cox model is based on the Analysis Data Model (ADaM) principles. The dataset is expected to have
+#'    one record per subject per analysis parameter. Rows in which the analysis variable (AVAL) or the sensor variable (CNSR) contain NA, are removed during analysis.
+#' @param equation Character vector, representing the covariates used at the Cox model. When NULL, the baseline (null) model is fit.
```
SHAESEN2, line 9: #' @param equation Character vector, representing the formula to be estimated with the Cox model. When NULL, the baseline (null) model is fit.

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+  ,AVAL = "AVAL"
+  ,conf.int = 0.95
+  ,...
+){
+
+  # Capture input to validate user input for data argument -----------------
+
+  ## Get actual data name as symbol
+  ### Magrittre pipe returns "." which inactivates recalls to survfit in downstream functions
+  ### map passes .x as cox_ccall$data
```
SHAESEN2, line 72: ### map passes .x as cox_call$data
  
  There was a typo

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+  if (!(inherits(data, "data.frame") | inherits(data, "tibble") | inherits(data, "data.table"))) {
+    stop(paste0("data can be of class `data.frame` or `tibble` or `data.table`."))
+  }
+
+  data <- as.data.frame(data)
+
+  # Validate columns --------------------------------------------------------
+
+  reqcols <- c(equation, CNSR, AVAL)
+  ss <- c("strata", "tt", "frailty", "ridge", "pspline", "cluster", "I", "factor")
```
SHAESEN2, line 98: I dont like following code because the function will evolve and some arguments in coxph will get depricated.

I would replace this part with something more defensive or don't overengineer it:
available <- names(as.list(args(survival::coxph)))

requested <- names(as.list(args(survival::survfit)))

intersect(requested, available)

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+  data <- as.data.frame(data) %>%
+    tidyr::drop_na(AVAL, CNSR)
+
+
+  # Ensure the presence of at least one strata -----------------------------
+
+  if (is.null(equation)) {
+    equation <- "1"
+  } else {
+    equation <- paste(equation, collapse = " + ")
```
SHAESEN2, line 132: For the equation, I would let the user fill it out completely. I found it very confusing to understand the examples untill I saw the function collapses it. 
For me it does not make sense for the cox. If I have to fill out an equation, I dont want to think about how it is interpreted by the downstream function. Therefore, I would remove the following code:

  } else {
    equation <- paste(equation, collapse = " + ")
  }

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+
+  ## Reverse censoring: see ADaM guidelines versus R survival KM analysis
+  library(survival)
```
SHAESEN2, line 137: remove line of code

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+    names(dots)
+    for (i in seq_along(dots)) {
+      survfit_object$call[[names(dots)[i]]] <- unlist(dots[i], use.names = FALSE)
+    }
+  }
+
+
+  # Add additional metadata -------------------------------------------------
+
+  if ("PARAM" %in% colnames(data) && length(setdiff(c("PARAMCD", "PARAM"), equation)) == 2) {
```
SHAESEN2, line 161: revise based on other comments

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+    # we expect only one unique value => catch mistakes
+    survfit_object[["PARAM"]] <- paste(unique(data[["PARAM"]]), collapse = ", ")
+  }
+
+  if ("PARAMCD" %in% colnames(data) && length(setdiff(c("PARAMCD", "PARAM"), equation)) == 2) {
+    # we expect only one unique value => catch mistakes
+    survfit_object[["PARAMCD"]] <- paste(unique(data[["PARAMCD"]]), collapse = ", ")
+  }
+
```
SHAESEN2, line 170: Dont we need to add 'Overall'/Null Model as strata when we dont fill in equation? When we plot the null model this needs to be clear that this is the null model as described in your documentation

R/estimate_cox.R
```diff
@@ -0,0 +1,181 @@
...
+
+  # Return ------------------------------------------------------------------
+
+
+  summary_cox <- summary(survfit_object, conf.int = conf.int)
```
SHAESEN2, line 175: We are creating a wrapper around coxph so you need to return the same object as coxph does, not a summary thereoff. We are allowed to extend the coxph object so if you need extra info from the summary, then add it to the coxph object that you return and document it as such.

SHAESEN2 (changes requested):

R/estimate_cox.R
AlexandraP-21: Done

R/estimate_cox.R
AlexandraP-21: Done

SHAESEN2:
add in reexports.R file export from survival STRATA blah

SHAESEN2:
see #362

Issue #261, opened 2021-09-14
`visR::add_risktable()` fails with a cryptic error message when the `strata` variable contains missing values.


gdario:
#### Issue description

When generating a Kaplan-Meier plot with a stratifying variable, `visR::add_risktable()` fails with a rather cryptic error message. If one does not include the risk table, the plot is generated without problems. It is only the creating of the risk table that produces the error, as illustrated in the example below

#### Steps to reproduce the issue

Please share code if possible.

```r
library(visR)
set.seed(42)

# Dataset without missing values
dataset <- data.frame(
  AVAL = rexp(n = 30, rate = 10),
  CNSR = sample(c(0, 1), size = 30, replace = TRUE),
  strata = gl(n = 2, k = 15, labels = c("A", "B"))
)

# Dataset with missing values
dataset_na <- dataset
is.na(dataset_na$strata[sample(seq_len(30), size = 5)]) <- TRUE

# Works
dataset %>%
  visR::estimate_KM(strata = "strata") %>%
  visR::visr(fun = "surv") %>%
  visR::add_risktable(label = c("Subjects at Risk", "Censored"),
                      statlist = c("n.risk", "n.censor"))

# Fails
dataset_na %>%
  visR::estimate_KM(strata = "strata") %>%
  visR::visr(fun = "surv") %>%
  visR::add_risktable(label = c("Subjects at Risk", "Censored"),
                      statlist = c("n.risk", "n.censor"))
```

#### What's the expected result?

We would expect the risk table to be added to the plot.

#### What's the actual result?

The following error message is produced:

```
Error: Tibble columns must have compatible sizes.
* Size 5: Column `na.action`.
* Size 25: Columns `time`, `n.risk`, `n.event`, `n.censor`, `surv`, and 9 more.
ℹ Only values of size one are recycled.
Run `rlang::last_error()` to see where the error occurred.
```

#### Additional details / screenshot

The issue is localised in the `tidyme()` file, and more precisely starting from [this line](https://github.com/openpharma/visR/blob/a70cc2d775f7510dd1f03edf77cac2421943063a/R/tidyme.R#L66). The missing values cause the objects that should be bound together to have different lengths, and this is what seems to be igniting the issue.

timtreis:
Hey @gdario! 

Thanks for pointing this out, I was able to reproduce the issue. We will fix this in the upcoming release and then ping you.

PR #271, opened 2021-10-12
Fixed issue - visR::add_risktable() fails with a cryptic error message ...

SHAESEN2:
Fix for issue #261


SHAESEN2:
@timtreis I believe you updated the testing file already. Thanks for that.

Issue #280, opened 2021-10-15
Test TOCs and las x'ed by statements in test headers aren't up to date


timtreis:
"Last updated by" and "Last update date" are rarely updated. Also, the TOC of tests at the top of our test files is manually copy/pasted and therefore usually async. This is unnecessary manual work and should be automated.

PR #281, opened 2021-10-15
Removed last x'd by/when comments

timtreis:
https://github.com/openpharma/visR/issues/280

### DIFF
--- a/tests/testthat/test-CRAN_watchdog.R
+++ b/tests/testthat/test-CRAN_watchdog.R
@@ -1,8 +1,4 @@
 #' @title Specifications CRAN_watchdog
-#' @section Last updated by:
-#' Tim Treis
-#' @section Last update date:
-#' 06-JULY-2021
 
 # Specifications ---------------------------------------------------------------
 #' T1. Our codebase doesn't violate CRAN style-guidelines.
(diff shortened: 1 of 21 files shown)


timtreis:
If these comments are used for anything in terms of verification, we definitely need a better system that hoping that everyone remembers to update them. Some files don't have them at all, some are definitely outdated (mine included). 

I'll probably implement something like our CRAN-watchdog for the test TOC at the top of these files anyway. I noticed that they're out of sync with the actual tests a lot of times which also seems less than ideal, so I'll just generate them by scraping the actual code. (That's what I wanted the `get_visR_files()` function for). I could probably also write something like this for updating these two fields based on `git blame`. But I think our current system doesn't work.

SHAESEN2:
The system only works if everyone is consistent with it. Currently this is the recommended practice.
Once you have established a better workflow, we can remove the headers. Can you please create an issue for updating the test headers and close this PR?

timtreis:
Can you document these verification procedures and their implied requirements for `visR` in the wiki? In the core meeting, no one was aware of them which is why we said that we could just get rid of the headers. Then I'll come up with a solution that is more robust than hoping that everyone remembers these.

SHAESEN2:
https://github.com/thebioengineer/validation

Issue #272, opened 2021-10-14
add_risktable y label gets moved 


SHAESEN2:
When adding the risktable to the plot, the y label moves entirely to the left of the plot, leaving a big gap between the Y-label and the plot. This should be adressed.

SHAESEN2:
Hey Tim,

You only see it when the labels are longer.

``` r
library(visR)

surv <- estimate_KM(adtte, strata=c("SEX", "TRTP"))

visr(surv) %>%
  add_risktable(group="statlist")
  
```

![image](https://user-images.githubusercontent.com/47894155/139635588-ec4ef90d-43f5-4f95-88e8-802f67710b49.png)

timtreis:
Thanks Steven, could reproduce!

Debugging WIP:
Issue most likely caused because `visR:::align_plots()` aligns the gtable with the x-axis positions. The strata-names are formally tick labels? Therefore, the y-label of the upper plot should be to the left of the tick labels and gets moved.

timtreis:
I think I managed to fix it, do you want to try to break it @SHAESEN2?

![image](https://user-images.githubusercontent.com/12818242/139680563-2f90abb3-2baf-4a2a-a1be-cc4663e1073d.png)

timtreis:
How relevant is this issue? The only actually robust solutions I see here are massively overengineered for what little benefit it brings. By fixing this, we're violating the underlying grid principle since we'd want to place a component outside of his designated square. This is calculated on the fly while rendering (afaik) so it'd be a gigantic mess (worth a lot of hours of code) to even find a somewhat robust solution.

SHAESEN2:
It should not be that difficult. Cowplot allows the label to be aligned by default between two plots or even allows it to be shifted to the most left side of the complete figure. Could you check their solution and see if we can implement similarly?

timtreis:
Don't really want to. I already poured 6+ hours into this and would rather get the table two and the auto-deploy of our interactive vignette to shinyapps.io running :D Maybe I'll check in with this later

SHAESEN2:
Understood. I will reassign the issue to myself.

timtreis:
Thanks!

SHAESEN2:
#294

SHAESEN2:
Closed with #294

PR #286, opened 2021-11-01
Moves y-label back to axis after `add_risktabel()`

timtreis:
As discussed in https://github.com/openpharma/visR/issues/272


NAMESPACE
```diff
@@ -26,12 +26,12 @@ S3method(visr,attrition)
 S3method(visr,default)
 S3method(visr,survfit)
 export("%>%")
+export(.align_plots)
```
SHAESEN2, line 29: keep exported as this function is very useful for other graphs and can be used out of the box

R/add_risktable.R
```diff
@@ -155,11 +155,8 @@ add_risktable.ggsurvfit <- function(
 
 # Align plot and table by adjusting width ---------------------------------
 
-  gglist <- list(gg) %>%
-    base::append(tbls)
-  
-  ggA <-  gglist %>%
-    align_plots()
+  gglist <- list(gg) %>% base::append(tbls)
+  ggA <-  gglist %>% .align_plots()
```
SHAESEN2, line 159: %>% then new line is the standard notation right?

R/utils_visr.R
```diff
@@ -24,12 +24,12 @@
 #' cowplot::plot_grid(plotlist = list(p1,p2), align = "none", nrow=2)
 #'
 #' ## align_plots() takes into account legend width
-#' cowplot::plot_grid(plotlist = align_plots(pltlist = list(p1, p2)), align = "none", nrow=2)
+#' cowplot::plot_grid(plotlist = .align_plots(pltlist = list(p1, p2)), align = "none", nrow=2)
```
SHAESEN2, line 27: align ="v", axis = "l", ncol = 1 to make explicit what we choose as default

R/utils_visr.R
```diff
@@ -61,6 +60,10 @@ align_plots <- function(pltlist) {
     x$widths <- max.widths
     x
   })
+  
+  # Move y-axis label back closer to the axis
+  strata_tick_width <- grid::convertUnit(plots.grobs.eq.widths[[2]]$grobs[[3]]$width, "cm")
+  plots.grobs.eq.widths[[1]]$grobs[[13]]$children[[1]]$x <- strata_tick_width - grid::unit(0.85, "cm")
```
SHAESEN2, line 66: Here I am worried. Why 0.85? is this emperical? If yes we need a more robust solution. You can check how cowplot is doing this when setting argument axis = "l".

tests/testthat/test-utils_visR.R
```diff
@@ -4,11 +4,11 @@
 
 # Specifications ---------------------------------------------------------------
 
-#' T1. `align_plots()` accepts a list of `ggplot` objects.
+#' T1. `.align_plots()` accepts a list of `ggplot` objects.
```
SHAESEN2, line 7: update to reflect the exported function. I think we should remain general eg The function accepts ...

tests/testthat/test-utils_visR.R
```diff
@@ -4,11 +4,11 @@
...
 #' T1.1 No error when a list of `ggplot` objects is passed.
 #' T1.2 An error when nothing or `NULL` is passed.
 #' T1.3 An error when a list containing non-`ggplot` objects is passed.
-#' T2. `align_plots()` aligns multiple `ggplot` objects, taking the legend into account.
+#' T2. `.align_plots()` aligns multiple `ggplot` objects, taking the legend into account.
 #' T2.1 No error when a list of ggplots is passed.
 #' T3. The function `legendopts()` translates the input to a `ggplot2`-compatible list.
```
SHAESEN2, line 13: We need a test that checks if the label is not aligned across the plots, but keeps with the individual plots. Perhaps we can make an argument out of it on the longer term similar to cowplot

SHAESEN2 (changes requested):
Nice work!

NAMESPACE
timtreis: It's still exported, no? I just gave it the helper function notation

SHAESEN2: I want to keep it as a utility function from visR as I believe it has added value (so without the . in front)

R/utils_visr.R
timtreis: Reworking this actually exposed unexpected behaviour that would've otherwise gone unnoticed! Nice
image

SHAESEN2:
More general solution in #294

Issue #93, opened 2020-10-15
KM plot show quantiles


SHAESEN2:
Add drop lines to show quantiles of the KM distribution.

SHAESEN2:
Also allow to add horizontal line for median survival using a wrapper around geom_hline(yintercept=xx,  linetype = xx)

timtreis:
@SHAESEN2 I'll probably have some time during R/Pharma 2021 to implement this, doesn't feel like a lot of work. Do you already have a name in mind? `add_survival_quantiles()' with "half" as the default?

timtreis:
Unsorted ideas:
- lines hitting the x-axis in the colour of the strata?
- horizontal lines solid?

SHAESEN2:
add_quantiles and make an S3 method out of it => add_quantiles uses get_quantiles.
I like the dotted lines you show in the example and the line color as per strata.

SHAESEN2:
I think I prefer a default of dotted lines. I asked @AlexandraP-21  to look into it as well. She has a good eye for those things.

AlexandraP-21:
I took a look at the lines and like them dotted. However, I think that maybe (light) grey (and not black) would be better. The coloring ones I found them a bit confusing and distracting to be honest. Hope this helps!

PR #288, opened 2021-11-02
add_quantiles

timtreis:
as discussed in #93 

tagging @joanacmbarros since I can't tag her as a reviewer?


R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
+#' @title Add lines indicating the survival quantiles to a survival plot
```
SHAESEN2, line 1: Make more general as you are describing a method here.

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+#'
+#' @description Method to add lines indicating the specified quantiles of the survival curve.
```
SHAESEN2, line 3: for any plot right?

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+#' @rdname add_quantiles
+#'
+#' @export
+
+add_quantiles <- function(gg, ...){
+  UseMethod("add_quantiles", gg)
+}
+
+#' @param gg A ggplot created with visR
+#' @param quantiles y-axis values for which to draw the lines (default: 0.5)
```
SHAESEN2, line 38: Vector of quantiles to be displayed on the probability scale

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+#' @param ... other arguments passed on to the method to modify \code{\link[ggplot2]{geom_line}}
+#'
+#' @rdname add_quantiles
+#' @method add_quantiles ggsurvfit
+#' @export
+
+add_quantiles.ggsurvfit <- function(gg, 
+                                    quantiles = 0.5,
+                                    linetype = "mixed",
+                                    linecolour = "strata",
```
SHAESEN2, line 51: if i want black, do i put "black"? this is not clear. Perhaps you can specify that this needs to be a character representing ....

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+                                    alpha = 1, 
+                                    ...) {
+  
+  if (!is.numeric(quantiles)) {
```
SHAESEN2, line 55: check if entire vector is numeric: base::all

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+  }
+  
+  if ((alpha > 1) | (alpha < 0)) {
+    
+    warning("Invalid `alpha` argument, must be between 0 and 1. Setting it to 1.0.")
+    alpha <- 1.0
+    
+  }
+  
+  model <- eval(gg$data$call[[1]])
```
SHAESEN2, line 69: we will get errors if only strata if we dont add the overall strata manually see add_risktable

R/add_quantiles.R
```diff
@@ -0,0 +1,154 @@
...
+  
+  cols_to_pivot <- colnames(model_quantiles) %>%
+    setdiff(c("strata", "quantity"))
+  
+  model_quantiles_long <- model_quantiles %>% 
+    tidyr::pivot_longer(cols = cols_to_pivot) %>%
+    dplyr::rename(quantile = name) %>%
+    dplyr::mutate(group = paste(strata, quantile, sep = "_")) %>%
+    dplyr::rename(n = value) %>%
+    dplyr::mutate(surv = 1-(as.numeric(quantile))/100) %>%
```
SHAESEN2, line 83: this will depend on the scale of the graph. Perhaps you can pull out the Y axis range. You need to check if it is on the probability scale, cumulative hazard etc. Is this info stored in the graph?

tests/testthat/test-add_quantiles.R
```diff
@@ -0,0 +1,19 @@
+#' @title Specifications add_CI
```
SHAESEN2, line 1: update

tests/testthat/test-add_quantiles.R
```diff
@@ -0,0 +1,19 @@
...
+#' @section Last update date:
+#' 10-MAY-2021
+
+# Specifications ---------------------------------------------------------------
+
+# Different surv functions?
+
+# Requirement T1 ---------------------------------------------------------------
+
+context("add_quantiles - T1. No errors when confidence intervals are added to the plots.")
```
SHAESEN2, line 13: should be more elaborate

SHAESEN2 (changes requested):

R/add_quantiles.R
timtreis: fixed, good catch

R/add_quantiles.R
```diff
@@ -0,0 +1,167 @@
...
+#' @export
+
+add_quantiles <- function(gg, ...){
+  UseMethod("add_quantiles", gg)
+}
+
+#' @param gg A ggplot created with visR
+#' @param quantiles vector of quantiles to be displayed on the probability scale, default: 0.5
+#' @param linetype string indicating the linetype as described in the aesthetics of ggplot2 \code{\link[ggplot2]{geom_line}}, default: mixed (horizontal lines are solid, vertical ones are dashed)
+#' @param linecolour string indicating the linetype as described in the aesthetics of ggplot2 \code{\link[ggplot2]{geom_line}}, default: strata, (horizontal lines are grey50, vertical ones are the same colour as the respective strata)
```
SHAESEN2, line 40: default: strata , <- remove comma

R/add_quantiles.R
SHAESEN2: I dont see the fix

SHAESEN2 (changes requested):

tests/testthat/test-add_quantiles.R
```diff
@@ -10,10 +10,98 @@
 
 # Requirement T1 ---------------------------------------------------------------
 
-context("add_quantiles - T1. No errors when confidence intervals are added to the plots.")
+testthat::context("add_quantiles - T1. tbd.")
```
SHAESEN2, line 13: Should we make sure all tests are in place before merging?

SHAESEN2 (commented):
Fine for me but make sure that all testing is in place before merging.

R/add_quantiles.R
timtreis: Now it is

tests/testthat/test-add_quantiles.R
timtreis: Steven, you're so impatient :D Yeah, I'll pring you when all possible use cases are functioning and test coverage is 100 %

R/add_quantiles.R
SHAESEN2: only for survival? then why create a method?

R/visr.R
```diff
@@ -168,7 +168,7 @@ visr.survfit <- function(
 # Y-label ----------------------------------------------------------------------
 
   if (is.null(y_label) & is.character(fun)){
-    y_label <- base::switch(
+    y_label <- switch(
```
SHAESEN2, line 171: keep base::switch as discussed. We never know if switch will occur in other packages.

R/visr.R
```diff
@@ -183,18 +183,18 @@ visr.survfit <- function(
     stop("No Y label defined. No default label is available when `fun` is a function.")
   }
   
-  if (is.character(fun)){
-    .transfun <- base::switch(
+  if (is.character(fun)) {
+    .transfun <- switch(
```
SHAESEN2, line 187: keep base::switch

R/visr.R
```diff
@@ -235,35 +236,35 @@ visr.survfit <- function(
 
 # Obtain X-asis label ----------------------------------------------------------
 
-  if (base::is.null(x_label)){
+  if (is.null(x_label)) {
```
SHAESEN2, line 239: Here OK to remove base

R/visr.R
```diff
@@ -279,21 +280,21 @@ visr.survfit <- function(
...
       pct = pretty(c(0, 100), 5),
       logpct = pretty(c(0, 5), 5),
-      cumhaz =  pretty(round(c(ymin, ymax), 0), 5),
+      cumhaz = pretty(c(ymin, ymax), 5),
       stop("Unrecognized fun argument")
     )
   } else if (is.null(y_ticks) & is.function(fun)) {
     
-    y_ticks = pretty(round(c(ymin, ymax), 0), 5)
+    y_ticks = pretty(c(ymin, ymax), 5)
```
SHAESEN2, line 297: it needs to be rounded to avoid clutter on the y axis. I agree that pretty is not the best choice. Did you find alternatives?
You could determine the amount of digits and then limit them to amount-1

R/visr.R
```diff
@@ -316,6 +317,19 @@ visr.survfit <- function(
     ggplot2::theme(legend.key = ggplot2::element_blank()) +
     NULL
   
+  # Save applied function so that we don't have to guess later on
+  if (is.primitive(fun)) {
+    # Prevent special case of log being a .Primitive from breaking add_quantiles
+    primitive_call <- utils::capture.output(fun)
```
SHAESEN2, line 323: will we have this as well with other primitive functions? Imagine I use log(-log) which is a common transformation in survival estimation?

tests/testthat/test-add_quantiles.R
```diff
@@ -0,0 +1,226 @@
...
+#' T1.4 An error when quantiles is NULL or not a numeric.
+#' T1.5 No error when `linetype` is a non-default character string.
+#' T1.6 No error when `linetype` is `mixed`.
+#' T1.7 An error when `linetype` is NULL or not a string.
+#' T1.8 No error when `linecolour` is a non-default character string.
+#' T1.9 No error when `linecolour` is `strata`.
+#' T1.10 An error when `linecolour` is NULL or not a string.
+#' T2. The quantile lines are y-axis-transformation dependent.
+#' T2.1 No error when the default `surv` option is used.
+#' T2.2 No error when any other implemented function is used.
```
SHAESEN2, line 20: A test for log or other primitive functions

SHAESEN2 (changes requested):

R/add_quantiles.R
timtreis: Right now we only have survival implemented, but maybe we'll add others? So I stuck with the structure. Also, I copied the file from `add_CI` and then just adjusted from there on. Theoretically we're not yet using that for anything else yet, but who knows.

R/visr.R
timtreis: In the end, `pretty` wasn't really the problem since that respects the value range. But, given that we were rounding with `0` precision, this would for example round 1.499999 to 1. 

The alternative approach would be a significantly more elaborate helper function that 1) rounds, 2) pretties and 3) then extend the upper/lower bound my extrapolating with the `pretty` result spacing between numbers to potentially add a 6th/7th before and/or after.

R/visr.R
timtreis: True, that currently breaks. I'll fix that

timtreis: fixed it

timtreis:
Check is at 0/0/0, codecov at 100 %

R/add_quantiles.R
SHAESEN2: A method should be general so please describe it generally

SHAESEN2: update to keep general.

R/visr.R
```diff
@@ -316,6 +317,21 @@ visr.survfit <- function(
     ggplot2::theme(legend.key = ggplot2::element_blank()) +
     NULL
   
+  # Save applied function so that we don't have to guess later on
```
SHAESEN2, line 320: This is really cool and we should highlight this in the documentation as well. Survminer is taking the lazy route:

  if(is.null(fun)) draw_lines <- TRUE
  else if(fun %in% c("cumhaz", "cloglog")){
    warning("Adding survival median lines is not allowed when fun is: ", fun)
    draw_lines <- FALSE
  }
  else if(fun == "pct") med_y <- 50

tests/testthat/test-add_quantiles.R
```diff
@@ -0,0 +1,281 @@
...
+#' T1.4 No error when `linetype` is a non-default character string.
+#' T1.5 No error when `linetype` is `mixed`.
+#' T1.6 No error when `linecolour` is a non-default character string.
+#' T1.7 No error when `linecolour` is `strata`.
+#' T2. The quantile lines are y-axis-transformation dependent.
+#' T2.1 No error when the default `surv` option is used.
+#' T2.2 No error when a function is passed as a string.
+#' T2.3 No error when a `.Primitive` function is used.
+#' T2.4 No error when a custom function is used.
+#' T3. If any parameter is specified incorrectly, defaults are used instead.
```
SHAESEN2, line 20: This seems to be a repeat of earlier tests. I like this general T3. and then putting everything below it regarding parameter choices.

Use argument instead of parameter

SHAESEN2 (changes requested):

SHAESEN2:
Nice work. Some details pending :)

SHAESEN2:
Hey @timtreis I have reviewed the branch. For me OK to merge now unless you have comments on my commit.

PR #317, opened 2022-02-06
Compatibility with broom-dependent workflows and increase traceability

SHAESEN2:
Included broom colname nomenclature in tidyme output
Included survfit object to tidyme output as attribute


ddsjoberg:
Hello! This is looking great!! 

1. The `estimate`, `conf.lower`, and `conf.upper` names need to be updated in `visr()` and `add_CI()` as well.
2. Need to update to remove this NOTE in the R CMD Check.
```r
N  checking R code for possible problems (8.6s)
   tidyme.survfit: no visible binding for global variable 'std.err'
   Undefined global functions or variables:
     std.err
```

SHAESEN2:
> Hello! This is looking great!!
> 
> 1. The `estimate`, `conf.lower`, and `conf.upper` names need to be updated in `visr()` and `add_CI()` as well.
> 2. Need to update to remove this NOTE in the R CMD Check.
> 
> ```r
> N  checking R code for possible problems (8.6s)
>    tidyme.survfit: no visible binding for global variable 'std.err'
>    Undefined global functions or variables:
>      std.err
> ```

Thanks for the review. I updated the naming in visR although I dont really see the point. Perhaps you missed that we store the potential transformations of the survival in est, est.lower and est.upper? For add_CI I dont see where there is an update required. Could you be more specific?

SHAESEN2:
Hey, it seems we indeed had something else in mind.
I thought you wanted to align the tidyme of visR with broom::tidy with respect to column names. This has been done.

During the plotting, we are allowing transformation, but I am not a fan of overwriting variable content (estimate, conf.high, conf.low) when doing a transformation and therefore we settled on est, est.lower and est.upper.

In your tidyr you need to add the Overall strata so I hope those three extra lines of renaming variables is acceptable? Are you sure users wont be interested in doing transformations on the data for the cum.inc.?

That being said, I fixed a bug in the visR code. I am a bit surprised our UAT did not pick it up.

ddsjoberg:
Hello @SHAESEN2 , I think the approach you've taken is good, particularly because there are many transformations users may want to make with the survfit object.

timtreis:
image

The test 4.1 in `test-tidyme.R` is doing a three-way comparison. This is not supported by `testthat` and can lead to false positives. I'll rewrite it, just wanted to also post it here for documentation.

timtreis (approved):
Works as expected and checks with 0/0/0

PR #324, opened 2022-02-25
Adding R CMD Checks for older versions of R

### LINKED ISSUE #323: Update GH Actions to Check R versions back to 3.5
- [ ] Update GH Actions to Check R versions back to 3.5 `usethis::use_tidy_github_actions()`
- [ ] Update DESCIRPITON file to depend on R >= 3.5

I think this should wait until after it's been decided to use just the main branch as the dev branch. If the current two-dev branch system remains, we'll need to modify the workflow files slightly to include all dev branches

ddsjoberg:
* Added the standard "tidy" workflow files including jobs to run R CMD Check on older versions of R and on multiple OS and build the pkgdown site when the main branch is updated. (#323)
* Increased minimum R version to 3.5 (#323)
* Updated `base::plot()` reference to `graphics::plot()`

The checks are failing on the older R versions. Need to investigate how to update the code to make it compatible with older R versions

closes #323


ddsjoberg:
@SHAESEN2 @timtreis I ran these tests on my local R 3.6 and it passes without error. When you have the chance, can you test if it runs for you locally without error? Thanks!

SHAESEN2:
Not sure how to check it. I did check package and it ran for most of it. Only the test for cuminc did not work. Joanna is working on this.

timtreis:
I could also directly make it a GH action, probably the more robust solution than our collective machines, no?

ddsjoberg:
Definitely GH Actions is the way to go. I added a workflow for the older versions of R back to 3.5. The R CMD Checks pass for recent releases, but fail for 3.6 and 3.5.

But I am not getting the error on my 3.6 R installation. Can to you @timtreis  check locally on your machine? Hopefully will get us closer to resolving?

ddsjoberg:
> Sure will do 👍

Thanks! My suspicion is that there is something in the testing code not compatible with older R versions, rather than the package code not being compatible. That section of the test file looked more complex

timtreis:
image

image

timtreis:
@ddsjoberg 

> That section of the test file looked more complex

What test code? The failure seems to be OS-dependent. I can also run it on my Ubuntu later

ddsjoberg:
The R CMD Check pass on recent R version on Ubuntu, and only fails on R 3.5 and 3.5
image

When I was looking on my machine, the errors were in an area of the testing code using `system()` to call and construct git commands.

timtreis:
These git commands are used to automatically polish our testfiles. However, I wrote all of that with running that locally in mind. Also, since this isn't run as a GH action anyway, the results that are generated during the R CMD check that runs on GitHub aren't even committed back to the repo. An easy fix would be to just put a "not-on-CRAN" tag into the test?

timtreis:
image
Once that one is through, I'll approve :)

timtreis:
image
Ready to squash and merge 🥳

timtreis (approved):
Works!

Issue #325, opened 2022-02-26
Create a contributor guide


SHAESEN2:
## Idea

**What is your idea for a new plot/table?**

**Do you have any examples you can share of what it looks like?**

**What potential packages/functions could be used to make this plot?**

**Are you interested to implement this yourself?**

SHAESEN2:
Updating the wiki might be easier.

ddsjoberg:
Update with details
- tibbles vs data.frames
- use of `$` internally
- details on the unit testing framework that has been implemented (sometimes when I do a full check of the package, the entire test file is deleted and re-saved with some changes to the header, I think). As this is a non-standard workflow, details for contributors would be incredibly useful
- add details for expectations of testing for new PRs. the package has amazing testing coverage, but as a new contributor is was difficult to orient myself on how to integrate my own unit tests

SHAESEN2:
Updated Wiki.

cschaerfe:
Updated https://github.com/openpharma/visR/wiki/Contribute with link to https://github.com/openpharma/visR/wiki/Coding-principles

PR #326, opened 2022-02-26
Create contributing

SHAESEN2:
adress issue #325 #326


ddsjoberg:
A contributing guide will be great.

1. If you keep the CRAN production, main stable dev, and develop branch as develop framework, add details about that here.
2. Depending on what you decide to do with the unit testing (ie. if you keep the current structure), please add more details about them here. For example, sometimes when I run the 
3. The guide says the repo adheres to the tidyverse style guidelines, but I don't think that is necessarily true. I think it's a good thing to do, but the package should also be styled with the styler package. Probably best to do this when there are no PRs open, because it can change many many files.
4. The guide can also stipulate other coding items that are expected, e.g. no using `$`, no tibbles, etc. FYI I did see this in the wiki that explicitly says to use tibbles. https://github.com/openpharma/visR/wiki/Coding-principles

FYI I did see this in the wiki https://github.com/openpharma/visR/wiki/Contribute A `CONRIBUTING.md` is the most common place to find this information in a repo though.

bailliem (approved):

SHAESEN2:
@cschaerfe I updated this file. The wiki has the same information, but we are not good at keeping it up-to-date. I propose to keep the information in only one spot. I think the package is the best place.

bailliem (approved):

PR #328, opened 2022-02-26
Create issue_template

SHAESEN2:
#327


ddsjoberg:
Does this template add to the existing templates, or does it replace them?
image

SHAESEN2:
I did not know it would be replaced. Probably good idea to keep the original?

SHAESEN2:
Agreee

Issue #335, opened 2022-02-26
integrate learn


SHAESEN2:
We should integrate the tutorial inside development

SHAESEN2:
closed via #339

PR #336, opened 2022-02-26
added learn tutorial from Tim

SHAESEN2:
#335


timtreis:
Hey Steven, the Code behind this PR seems to be the default template of learnR, isn't it? I have a GH action for the autodeploy with the tutorial in a separate branch.

SHAESEN2:
> Hey Steven, some of the code behind this PR seems to be the default template of learnR, isn't it? I have a GH action for the autodeploy with the tutorial we gave at the past workshops in a separate branch (https://github.com/openpharma/visR/tree/shinyappstest). Since @ddsjoberg rightfully brought up that our webpage could use a little love, maybe this would be a good point in time to also think about that? We could split our vignettes into "static" and "interactive" and then also have the interactive ones auto-deploy.
> 
> Also, the interactive tutorial we have right now is also a broad overview of everything. I think it'd be nice to split it by topics or turn it into case-studies.

Hey Tim,
Yes it is the same as I dont like pending things floating around. Could we merge it into develop and then create an issue to deploy it on our website? I like your idea alot. It makes for a nice small project for a new member.

Issue #337, opened 2022-02-28
Riskmetric badge is not rendered correctly


timtreis:
![image](https://user-images.githubusercontent.com/12818242/155969854-abb25a86-716b-4657-84b5-515a75fa393d.png)

timtreis:
The badge looks weird because shilds.io expects an URL of the format <string>-<number>-<colour>. However, because our riskmetric score is now negative, that format is broken. A riskmetric score should be in [0,1] though, so this issue seems to be an issue of `riskmetric` ? At least I'm also getting negative scores on my local machine. I filed an issue with them for it -> https://github.com/pharmaR/riskmetric/issues/246

timtreis:
Looks like it's an issue of their latest 0.1.2 release. I've now changed our `remotes` install candidate to the previous 0.1.1 which gives expected output.

timtreis:
The `riskmetric` team intends to fix this in an upcoming release (https://github.com/pharmaR/riskmetric/pull/242). I'll leave this issue open as a reminder to revert the PR in which we now specify riskmetric@0.1.1 once the release candidate of `riskmetric` is functional again.

timtreis:
The PR that's supposed to fix the issue is now merged, I'll keep an eye on when the release will be.

PR #338, opened 2022-02-28
WIP for GH action -> Fixes riskmetric badge 

timtreis:
See #337

### DIFF
--- a/.github/workflows/rebuild-readme.yaml
+++ b/.github/workflows/rebuild-readme.yaml
@@ -24,7 +24,7 @@ jobs:
           remotes::install_cran("data.table")
           remotes::install_cran("covr")
           remotes::install_github("lorenzwalthert/gitsum")
-          remotes::install_github("pharmaR/riskmetric")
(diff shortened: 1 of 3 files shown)


timtreis:
Using `riskmetric` v0.1.1 restored our README to the expected output:

image

ddsjoberg:
@timtreis the riskmetrics package has a CRAN release of v0.1.2, and the workflow is installing v0.1.1 from GitHub. 

Does the riskmetrics pkg no longer support the functionality being used? If they no longer support the functionality, should it not just be removed from the README? If it's a bug in the recent release, would it not be better to file an issue and patch the bug and install the most recent GH version?

timtreis:
Yes, I intentionally set it to install v0.1.1 instead of the most recent release. While 0.1.1 correctly yields a value in [0,1], the 0.1.2 version gives something in the area of -1.5 which breaks the shields.io template and according to the riskmetric doc should be an invalid value anyway. 

I have already filed an issue with them, but I'd suggest to just use 0.1.1 until it is fixed.

PR #348, opened 2022-04-14
Updated the strata levels and label in `visr()` 

### LINKED ISSUE #343: `visr.survfit()` strata label update?
The default label in the KM curve includes the variable name. This labeling comes from the default output in `survfit()`. 

image

But, it's difficult to imagine a scenario where anyone would want the variable name prefix in the figure. Because we have total control of the inputs with `estimate_KM()`, we can pretty easily update this to only include the levels without the `"varname="` prefix. When there are multiple stratifying variables we can use the `interaction()` function to construct a more beautiful stratifying variable to include.

``` r
interaction(gtsummary::trial$stage, gtsummary::trial$grade, sep = ", ") |> levels()
#>  [1] "T1, I"   "T2, I"   "T3, I"   "T4, I"   "T1, II"  "T2, II"  "T3, II" 
#>  [8] "T4, II"  "T1, III" "T2, III" "T3, III" "T4, III"
```

<sup>Created on 2022-03-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Also, what do you think about making the default strata label a blank space instead of include `"strata"` above the levels for every plot?

ddsjoberg:
Updated the strata levels and label in `visr()` for KM and cumulative incidence objects. Rather than displaying "strata" as the label, we now print the variable label (if available) or the variable name. The levels no longer include the hard-coded print of the variable name.

Example of the updated figure below!

closes #343

``` r
library(visR)
estimate_KM(data = adtte, strata ="TRTP") |> visr()
```

<sup>Created on 2022-04-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/add_CNSR.R
```diff
@@ -103,7 +103,8 @@ add_CNSR.ggsurvfit <- function(gg, shape = 3, size = 2, ...){
                                      y = est,
                                      color = strata),
                         shape = shape,
-                        size = size)
+                        size = size,
+                        show.legend = FALSE)
```
SHAESEN2, line 107: We decided not to include this on purpose. The coloring should be sufficient without showing the censoring symbols. If you have a specific business need, we can reopen the discussion.

R/tidyme.R
```diff
@@ -87,9 +87,16 @@ tidyme.survfit <- function(x, ...) {
       retme[["n.strata"]] <- rep(x[["n"]], x[["strata"]])
     }
   }
-  
+
   attr(retme, "survfit_object") <- survfit_object
+  # modify strata label, removing ref to raw variable name
+  if (!is.null(purrr::pluck(survfit_object, "estimate_KM_args", "strata"))) {
```
SHAESEN2, line 93: can we avoid the dependency on purrr? I dont think we are using it somewhere else.

R/tidyme.R
```diff
@@ -87,9 +87,16 @@ tidyme.survfit <- function(x, ...) {
...
+    for (stratum in purrr::pluck(survfit_object, "estimate_KM_args", "strata")) {
+      retme[["strata"]] <- stringr::str_remove(retme[["strata"]], stringr::fixed(paste0(stratum, "=")))
```
SHAESEN2, line 95: We decided not to depend on stringr. BASE R should be sufficient for sting manipulation.

R/visr.R
```diff
@@ -560,3 +562,11 @@ visr.tidycuminc <- function(x = NULL
   gg
 }
 
+
+.construct_strata_label <- function(estimate_args, sep = ", ") {
+  if (is.null(estimate_args[["strata"]])) return(" ")
```
SHAESEN2, line 567: Do we expect this to be NULL?
If we estimate using the formulat ~ 1 we add an artifical stratum "Overall" to facilitate downsteam processing.

R/estimate_KM.R
```diff
@@ -72,13 +72,14 @@ estimate_KM <- function(
 ){
 
 # Capture input to validate user input for data argument -----------------
+  dots <- rlang::dots_list(...)
+  estimate_KM_args <- as.list(environment())
```
SHAESEN2, line 76: I would only keep what we actually need from the environment() call. In clinical trials we work with big data so we don't want to keep information that is not required eg estimate_KM_args[["data"]]. If the strata is NULL is should be defaulted to "Overall" with label "Overall"

SHAESEN2: I'm not sure if I see the point of this additional line since we know the strata and we can remove whatever is in front of 'xx = ' easily + look for the corresponding label in the dataframe. Perhaps another argument in survfit object that contains strata labels might be more useful?

R/estimate_KM.R
```diff
@@ -72,13 +72,14 @@ estimate_KM <- function(
 ){
```
SHAESEN2, line 72: Can you document the added behaviour to the roxygen section? It would facilitate review as I am not quite sure about the implemented logic although I know the endgame.

R/estimate_KM.R
```diff
@@ -146,8 +147,9 @@ estimate_KM <- function(
 
 
 # Update Call with original info and dots, similar as update.default ------
-  
-  survfit_object$call[[1]] <- quote(survival::survfit)
+
+  survfit_object[["estimate_KM_args"]] <- estimate_KM_args
```
SHAESEN2, line 151: adding the entire list of information seems redundant. We have all this information already in the survfit object.

R/estimate_KM.R
SHAESEN2: What I was expecting here is that you are capturing the actual environment of 'data' so that we don't get in trouble with wrapper functions.

R/estimate_KM.R
SHAESEN2: Once the logic is agreed upon, we need to add UAT for the PR. I can assist with this if required.

SHAESEN2 (changes requested):

R/add_CNSR.R
ddsjoberg: I am not sure I understand this comment. After my other updates, the CNSR symbol legend was being added rather than integrated with the line color legend. I agree the color legend should be sufficient, which is now what shows.

R/tidyme.R
ddsjoberg: gone!

R/tidyme.R
ddsjoberg: gone!

R/visr.R
ddsjoberg: i updated the structure, so this line no longer appears. In this setting, the strata arg was NULL

R/estimate_KM.R
ddsjoberg: now just exporting the strata variables as a named list that include the labels (if they are present). Add the env, I feel, should be addressed in another PR.

R/estimate_KM.R
ddsjoberg: updated to just return the strata label

R/estimate_KM.R
ddsjoberg: I added a note about the strata list that is now returned with the object.

ddsjoberg:
Thanks for the review @SHAESEN2 ! I think the updates def make the changes better.
- there are already unit tests in place for assessing the correct levels in the `visr()` legend
- can you write a unit test for testing the legend title? I am not familiar with the internals of ggplot objects, and don't know how to  construct the test. I think 2 tests: one without strata and one with strata would be good. When there are no strata, the title is `" "`, otherwise, the legend title is the variable name/label

R/estimate_KM.R
```diff
@@ -11,6 +11,7 @@
 #'      \item{The cumulative hazard is estimated using the Nelson-Aalen estimator (ctype = 1): H.tilde = cumsum(x$n.event/x$n.risk).
 #'      The MLE (H.hat(t) = -log(S.hat(t))) can't be requested.}
 #'      \item{A two-sided pointwise 0.95 confidence interval is estimated using a log transformation (conf.type = "log").}
+#'      \item{A named list of strata (and their lables), if present}
```
SHAESEN2, line 14: This item list are things to be aware of, not the content of the survival object. I propose you rephrase this or explain it somewhere else.

R/estimate_cuminc.R
```diff
@@ -38,6 +38,8 @@ estimate_cuminc <- function(data
                             ,...) {
   # check for installation of tidycmprsk package
   rlang::check_installed("tidycmprsk", version = "0.1.1")
+  dots <- rlang::dots_list(...)
+  estimate_cuminc_args <- as.list(environment())
```
SHAESEN2, line 42: I would implement the labels in a similar way here.

R/visr.R
```diff
@@ -560,3 +562,66 @@ visr.tidycuminc <- function(x = NULL
...
+    }
+    if (inherits(x, "tidycuminc")) {
+      data<- x$data
+    }
+
+
+
+    # get the original data frame
+    if (inherits(x, "survfit")) {
+      data <- as.list(x$call)[["data"]] %>% eval()
```
SHAESEN2, line 582: If we have the logic of strata labels in estimate_xx it would make it easier. You would do all the heavy lifting in a spot where all the information is available. Using this logic, you wont need to rebuild what you originally had.

The disadvantage is when we are passing a foreign survift object to visr(). The labels wont be adjusted.

Do we sacrifice efficiency or compatibility?

SHAESEN2 (changes requested):

R/estimate_cuminc.R
ddsjoberg: this line was left over from the last round of review. I removed the line that exports this object with the result, but forgot to delete the line created the object

R/estimate_KM.R
ddsjoberg: updated!

R/visr.R
ddsjoberg: I am not sure what happened...I thought I had deleted these lines. Anyway, they are now removed!

ddsjoberg:
Thanks @SHAESEN2 again for the review. I addressed all comments, and was able to add unit tests confirming the legend title is correct/as expected.

Can you take another look?

R/estimate_KM.R
```diff
@@ -5,6 +5,7 @@
 #'    argument are removed.
 #'    Alternatively, PARAM/PARAMCD can be used in the \code{strata} argument. \cr
 #'    The result is an object of class \code{survfit} which can be used in downstream functions and methods that rely on the \code{survfit} class.
+#'    When stratum are present, the returned survfit object is supplemented with the a named list of the stratum and associated label, if present.
```
SHAESEN2, line 8: Documentation is clear now: stratum -> strata

R/estimate_KM.R
```diff
@@ -184,6 +185,15 @@ estimate_KM <- function(
     }
   }
 
+  # add strata labels ----------------------------------------------------------
+  if (!is.null(strata)) {
+    survfit_object[["strata_lbls"]] <-
+      lapply(as.list(strata), function(x) attr(data[[x]], "label") %||% x) %>%
+      rlang::set_names(strata)
+  }
```
SHAESEN2, line 193: else if (is.null(strata){
  survfit_object[["strata_lbls"]] <- "Overall"
  names(survfit_object[["strata_lbls"]]) <- "Overall"
}

R/tidyme.R
```diff
@@ -87,9 +87,21 @@ tidyme.survfit <- function(x, ...) {
       retme[["n.strata"]] <- rep(x[["n"]], x[["strata"]])
     }
   }
-  
+
   attr(retme, "survfit_object") <- survfit_object
+  strata <- .extract_strata_varlist(survfit_object)
```
SHAESEN2, line 92: where is this function defined? I couldnt find it

tests/testthat/test-estimate_KM.R
```diff
@@ -22,6 +22,7 @@
 #' T3.3 When no strata are specified, an artificial strata is created 'Overall'
 #' T3.4 When only 1 stratum is specified, the stratum names are added to the `names` attribute'
```
SHAESEN2, line 23: We already had strata labels in the names attribute of survfit_object[["strata"]]. Can we not leverage that logic instead of adding another list to the survfit_object? Sorry to notice it this late.

For me its to include another element, but it seems overkill.

tests/testthat/test-estimate_KM.R
```diff
@@ -22,6 +22,7 @@
...
 #' T3.5 When more than 1 strata is specified, the stratum names are available in the `names` attribute
+#' T3.6 The strata label is constructed correctly
```
SHAESEN2, line 25: I'm not in favor of general descriptions. Can you explain what you mean with 'correctly'?
Perhaps: The strata labels are also available in a separate list element of the survfit object.

You also need to write a test to ensure:
- that this list element is added to the object.
- that the content is equal to the label of the corresponding variable in the data
- test if combination of labels is done as expected
- test if Overall is used if ~1 in formula

tests/testthat/test-tidyme.R
```diff
@@ -75,8 +75,9 @@ testthat::test_that("T3.2 The S3 method, associated with a `survfit` object, has
 
   survfit_object <- visR::estimate_KM(data = adtte, strata = "TRTA")
   survfit_object_tidy <- tidyme(survfit_object)
+  survfit_object$strata_lbls <- NULL
```
SHAESEN2, line 78: This is cheating :) You need to add the labels to surv_object_df to demonstrate that each list element is available in the final dataframe

SHAESEN2 (changes requested):

R/estimate_KM.R
SHAESEN2: After some spot checks, I noticed we also have the required information in the names attribute of the strata element. I'm OK with adding that information into a new list element, but it should be done consistently (same logic for the attribute and the list element)

ddsjoberg: > else if (is.null(strata){ survfit_object[["strata_lbls"]] <- "Overall" names(survfit_object[["strata_lbls"]]) <- "Overall" }

This is for the label, not the levels (which are already handled)

> After some spot checks, I noticed we also have the required information in the names attribute of the strata element. I'm OK with adding that information into a new list element, but it should be done consistently (same logic for the attribute and the list element)

That object has the raw levels of the stratifying variables, not the labels for a strata variable

R/tidyme.R
ddsjoberg: In RStudio, you can find any function object by pressing `Ctrl+.`

tests/testthat/test-estimate_KM.R
ddsjoberg: that object was not modified in this PR, and those are the strata levels

tests/testthat/test-estimate_KM.R
ddsjoberg: All of these are tested already, although not each an explicit separate test. 
- that this list element is added to the object: already testing equality to a list object, there it must be a list.
- that the content is equal to the label of the corresponding variable in the data: this is the test that already appears
- that the content is equal to the names attribute of the strata element of the survfit object: does not apply, those are strata levels, not the strata labels
- test if combination of labels is done as expected: tested in the visr file, where the strata label is checked
- test if Overall is used if ~1 in formula: also already in the visr test file

tests/testthat/test-tidyme.R
ddsjoberg: corrected

R/estimate_KM.R
ddsjoberg: updated.

ddsjoberg:
@SHAESEN2 comments addressed. if you have additional comments, perhaps we can jump on a call to resolve quickly.

R/estimate_KM.R
```diff
@@ -184,6 +185,15 @@ estimate_KM <- function(
     }
   }
 
+  # add strata labels ----------------------------------------------------------
```
SHAESEN2, line 188: Should not be a new section, but rather a comment.

R/estimate_KM.R
SHAESEN2: I think they can still be aligned correct?
We need to have "Overall" in the legend when no strata are specified

R/tidyme.R
```diff
@@ -22,8 +22,8 @@
 #' lm_tidied <- visR::tidyme(lm_object)
 #' lm_tidied
 #'
-#' @return Tibble containing all list elements of the S3 object as columns. 
-#'   The column 'strata' is a factor to ensure that the strata are sorted 
+#' @return Tibble containing all list elements of the S3 object as columns.
```
SHAESEN2, line 25: Dataframe instead of tibble?

R/tidyme.R
SHAESEN2: I know. I meant that there was no function defined in the files :)

SHAESEN2 (changes requested):

R/estimate_KM.R
SHAESEN2: > I think they can still be aligned correct? We need to have "Overall" in the legend when no strata are specified

SHAESEN2: Can you put a comment to explain the differences between the list element of labels and the attribute above in the code?

Also explain briefly why you would not add "Overall" here.

tests/testthat/test-estimate_CUMINC.R
```diff
@@ -6,6 +6,8 @@
 #' T1. The function wraps `tidycmprsk::cuminc()` appropriately
 #' T1.1 No errors `estimate_cuminc()`
 #' T1.2 No errors `estimate_cuminc()` sister functions
+#' T2. The legend follows the model strata labels
+#' T2.2 The color legend title is accurate.
```
SHAESEN2, line 10: accurate = represents the model strata label

tests/testthat/test-estimate_KM.R
SHAESEN2: @SHAESEN2 to look into this again in final review.

SHAESEN2 (changes requested):

tests/testthat/test-visr_survfit.R
```diff
@@ -43,6 +43,7 @@
 #' T3.7 The `fun` argument is stored in the final object as attribute `fun`.
 #' T4. The legend follows the model strata labels
 #' T4.1 The legend follows the model strata labels
```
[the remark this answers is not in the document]
ddsjoberg: more detail added

tests/testthat/test-visr_survfit.R
```diff
@@ -43,6 +43,7 @@
...
+#' T4.2 The color legend title is accurate.
```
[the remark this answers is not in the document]
ddsjoberg: more detail added

R/estimate_KM.R
ddsjoberg: if that is your preference, i'll remove

R/estimate_KM.R
ddsjoberg: more detail added in the comments

R/tidyme.R
ddsjoberg: moved to utils file

R/tidyme.R
ddsjoberg: i updated this text

SHAESEN2:
Amazing update and well tested!

PR #356, opened 2022-04-23
Added update_documentation to quickly update everything and generate …

### LINKED ISSUE #301: Documentation in S3 method `visr()`
Some arguments are documented for each class/method. When these arguments are documented multiple times, the last one is used in the help file.

For example, the `visr(x=)` argument is documented as 
- `@param x Object of class 'survfit'`
- `@param x Object of class 'attritiontable' with each row corresponding to an inclusion step in the cohort and minimally a description and a count column`

Only the last one is shown in the rendered help file.

SHAESEN2:

documentation/visR.pdf
Updated S3 method roxygen

**Which files are involved in this pull request and what changes were made?**
All

NA

closes #301 

Checklist for PR reviewer

- [x] PR branch has pulled the most recent updates from main branch. Ensure the pull request branch and your local version match and both have the latest updates from the main branch.
- [x] If a new function was added, function should be included in `_pkgdown.yml`
- [x] If a bug was fixed, a unit test was added for the bug check
- [x] Run `pkgdown::build_site()`. Check the R console for errors, and review the rendered website.
- [x] Code coverage is suitable for any new functions/features. Review coverage with `withr::with_envvar(new = c("NOT_CRAN" = "true"), covr::report())`. Before you run, begin a fresh R session without any packages loaded. 
- [x] R CMD Check runs without errors, warnings, and notes
- [x] `usethis::use_spell_check()` runs with no spelling errors in documentation
- [x] Has `NEWS.md` been updated with the changes from this pull request under the heading indicating the latest version. If there is an issue associated with the pull request, reference it in parentheses at the end update (see `NEWS.md` for examples).
- [x] Has the version number been incremented using `usethis::use_version(which = "dev")` 
- [x] Approve Pull Request


R/utils_general.R
```diff
@@ -14,3 +14,47 @@
...
+#' @examples
+#' \dontrun{
+#'   update_documentation(pkg = "~/visR")
+#'  }
+
+update_documentation <- function(pkg = "~/visR", fileNm="visR")
+  {
+
+    ## Create Rd files document ####
+    devtools::document(pkg = pkg)
```
ddsjoberg, line 43: FYI, this adds a {devtools} dependency (a pretty big one). If you choose to keep it, it needs to be added to DESCRIPTION

R/utils_general.R
```diff
@@ -14,3 +14,47 @@
            width = 6,
            scientific = FALSE)
 }
+
+#' @title Update documentation
```
ddsjoberg, line 18: Can you add some details describing the need for this function fulfills? 

I see it creates a PDF of the help files. Is this something needed given that the package build already includes the PDF? 

I personally find the website to be a much much much better representation of the help files compared to the PDFs.

ddsjoberg (changes requested):
1. You added a `documentation` folder. If you keep this folder, it needs to be added to `.Rbuildignore` to avoid a NOTE on r CMD Check. NOTES on R CMD Check will exclude the package from auto-acceptance on CRAN, and will kick it to a human for review.
2. FYI R 4.2 was released yesterday, and they have added an additional check for using `inherits(strata, "list")` instead of `class(strata) == "list"`, for example.
3. For generic functions like `visr()`, where the the generic and S3 methods are all documented in a single file–please move all documentation of arguments to the top of the file. This makes it much easier to review, reduces risk of argument doc duplication, and adds parity between the documentation code and the rendered help file (where all documented arguments all appear on top).

ddsjoberg:
@SHAESEN2 overall, the doc updates look great! 👍🏼 Added a few comments

R/utils_general.R
SHAESEN2: Also if it is only for internal usage?

R/utils_general.R
SHAESEN2: I have moved it locally and see if we get the request for some documentation in pdf inside the package.

R/utils_general.R
SHAESEN2: > 1. You added a `documentation` folder. If you keep this folder, it needs to be added to `.Rbuildignore` to avoid a NOTE on r CMD Check. NOTES on R CMD Check will exclude the package from auto-acceptance on CRAN, and will kick it to a human for review.
Thanks for the note. I did not realize it.

> 2. FYI R 4.2 was released yesterday, and they have added an additional check for using `inherits(strata, "list")` instead of `class(strata) == "list"`, for example. Perhaps not in this PR, but all these calls will need to be updated. While you're updating, note that `inherits(x, c("a", "b"))` is equivalent to `inherits(x, "a") | inherits(x, "b")`.
Will you make an issue for this? 

> 3. For generic functions like `visr()`, where the the generic and S3 methods are all documented in a single file–please move all documentation of arguments to the top of the file. This makes it much easier to review, reduces risk of argument doc duplication, and adds parity between the documentation code and the rendered help file (where all documented arguments all appear on top).
Agree and updated

R/utils_general.R
ddsjoberg: FYI
- CRAN builds and ships a PDF document with every release. https://cran.r-project.org/web/packages/visR/visR.pdf
- You can build that manual anytime you'd like with `devtools::build_manual()`

R/utils_general.R
ddsjoberg: > Also if it is only for internal usage? I have moved it to my local folder now.

Yes, because the code was in the R folder, and was shipped as part of the package, it would need to be listed as a dependency.

ddsjoberg (approved):
looks great!

PR #359, opened 2022-04-24
Updates to `class()` calls

### LINKED ISSUE #346: Order of the newly assigned classes
Quoting from the Advanced R book:

> Objects are not limited to a single class, and can have many classes:
> `class(x) <- c("A", "B")`
> `class(x) <- LETTERS`
> As discussed in the next section, R looks for methods in the order in which they appear in the class vector. So in this example, it would be like class A inherits from class B - if a method isn’t defined for A, it will fall back to B. However, if you switched the order of the classes, the opposite would be true!
http://adv-r.had.co.nz/S3.html

In visR, I noticed that some classes are defined in the opposite order: the most generic class is defined first and the specific visR classes are last. In the example below, we see that the ggplot classes are listed first, meaning that if ggplot2 were to ever implement a `visr()` method for class "ggplot" it would take precedent over the visR-defined method functions.

Generally, you want the most broad class listed last, and the most narrow and specific class should be listed first to ensure the S3 dispatch system selects the most appropriate method.  So the classes below should be defined as `c("ggsurvfit", "gg", "ggplot")`

``` r
library(visR)

adtte %>%
  estimate_KM(strata = "TRTP", conf.int = 0.90) %>%
  visr(legend_position = "right", x_unit = "Days") %>%
  class()
#> [1] "gg"        "ggplot"    "ggsurvfit"

adtte %>% 
  ggplot2::ggplot() %>% 
  class()
#> [1] "gg"     "ggplot"
```

<sup>Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

As an example, the Hmisc package adds the class "labelled" to labelled objects. The new labelled class is added to the beginning.

``` r
Hmisc::label(mtcars$mpg) <- "Miles per Gallon"
class(mtcars$mpg)
#> [1] "labelled" "numeric"
```

<sup>Created on 2022-03-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #358: R 4.2 Added check for no `class(x)==`
FYI R 4.2 was released on April 22, and they have added an additional check for using `inherits(strata, "list")` instead of `class(strata) == "list"`, for example. All these calls will need to be updated. While updating, note that `inherits(x, c("a", "b"))` is equivalent to `inherits(x, "a") | inherits(x, "b")`.

ddsjoberg:
- Removed calls like `class(x) == "list"` in accordance with the R 4.2 R CMD Checks (#358)
- Re-ordered the class assignment as needed (#346)

Unit tests already in place

closes #358 
closes #346

Checklist for PR reviewer

- [ ] Approve Pull Request


SHAESEN2 (approved):
I found more occurrences of class vs inherits. Please review.

SHAESEN2:
Thank you for taking this on !

ddsjoberg:
@SHAESEN2 thanks for the review! Good catch, I had only updated the calls in the package functions that get checked by R CMD Check. It's better practice to use this in the test files too.

PR #361, opened 2022-05-01
Review of Joanna testing

### LINKED ISSUE #299: Check if GT is still robust after the major releases they had
Daniel (gtsummary) mentioned that GT had a big release and made major changes to their API. We're requiring [GT 0.1.0](https://github.com/openpharma/visR/blob/d7929a026f354e3ebaae47837764e9035f62adcf/DESCRIPTION#L43), which was supposedly pre-CRAN?  maybe me should check if our old code still runs without problems.

@cschaerfe

SHAESEN2:
estimate_cuminc
--------------
- Initialize data argument
- focus on arguments that are required for ADaM eg confint can be defined through ... Tidycmprisk should deal with confint variations
- add more informative messages
- Do we want to add an overall strata? We can put Overall instead of 1 inside the formula and then add a column strat99= Overall
- %||% is not defined. When strata is NULL then the formula becomes ...... ~ NULL in the final object without any error => used ifelse statement

testing
-------
- we need to focus on what the wrapper is doing, not on what the tidycmprisk function
 -- test arguments
 -- test excecuting surrounding the main function
 -- do not test how the main function works
- downstream processing of this object needs to be tested in the respective test files, not in the estimate function
- we expect tidycmprisk to do the heavy lifting and testing on the final object. We only verify wether the wrapper does not change the object. If it does change the object, we need to demonstrate it does not influence the results and that the changes are actually present

closes #299


SHAESEN2:
@ddsjoberg do you agree with the updates?
@joanacmbarros Can you move your tests/create tests for the cuminc downstream processing? Please have a look at the updated file and the above comments.

ddsjoberg:
The updates look good to me. Just two comments.

> * focus on arguments that are required for ADaM eg confint can be defined through ... Tidycmprisk should deal with confint variations

Putting the `conf.int=` argument in `estimate_cuminc()` avoids an unnecessary double estimation of the entire object. Please add it back.

> * Do we want to add an overall strata? We can put Overall instead of 1 inside the formula and then add a column strat99= Overall

Please let's keep it as it is. It'll avoid downstream changes that would be required, and using the 1 in the formula is what `estimate_KM()` does as well.

R/estimate_cuminc.R
```diff
@@ -20,34 +20,60 @@
...
 #'
 #' @examples
-#' estimate_cuminc(
-#'   tidycmprsk::trial,
+#' visR::estimate_cuminc(
+#'   data = tidycmprsk::trial,
 #'   strata = "trt",
 #'   CNSR = "death_cr",
 #'   AVAL = "ttdeath"
-#' ) %>%
```
ddsjoberg, line 28: Please keep the full example. The generic functions (e.g. `visr()`, `add_CI()`, etc) don't have cuminc examples, and this is the only place the user will see them in use in the documentation.

ddsjoberg:
I just pushed a couple changes FYI
1. There was an invalid `url{}` tag in `align_plots()` help file that spanned two rows, and could not be parsed by pkgdown. I fixed it
2. I updated the example in `estimate_cuminc()` to print the output from `estimate_cuminc()`, then print the visualizaton.
3. Added the `conf.int=` argument to `estimate_cuminc()`

ddsjoberg:
> - %||% is not defined. When strata is NULL then the formula becomes ...... ~ NULL in the final object without any error => used ifelse statement

It's fine if you prefer not to use it, but it is defined and imported from {rlang}

R/estimate_cuminc.R
SHAESEN2: We will add examples while finishing the test files for the respective functions. Is that OK?

R/estimate_cuminc.R
```diff
@@ -20,29 +20,64 @@
...
+#' cuminc <-
+#'   visR::estimate_cuminc(
+#'     data = tidycmprsk::trial,
+#'     strata = "trt",
+#'     CNSR = "death_cr",
+#'     AVAL = "ttdeath"
+#'   )
+#' cuminc
+#'
+#' cuminc %>%
```
SHAESEN2, line 32: These should migrate to the respective functions

R/estimate_cuminc.R
```diff
@@ -20,29 +20,64 @@
...
+#'   visR::visr() %>%
+#'   visR::add_CI() %>%
+#'   visR::add_risktable(statlist = c("n.risk", "cum.event"))
+
+estimate_cuminc <- function( data = NULL
                             ,strata = NULL
                             ,CNSR = "CNSR"
                             ,AVAL = "AVAL"
                             ,conf.int = 0.95
```
SHAESEN2, line 41: Based on your feedback, I assume there is no default CI estimation in tidycmprsk? If you this is the case, then you need to a test for CI argument: (numeric, not numeric, when you change, the estimation changes, checks for 0<x<1

SHAESEN2:
> The updates look good to me. Just two comments.
> 
> > * focus on arguments that are required for ADaM eg confint can be defined through ... Tidycmprisk should deal with confint variations
> 
> Putting the `conf.int=` argument in `estimate_cuminc()` avoids an unnecessary double estimation of the entire object. Please add it back. The returned object from this function includes CIs and the users need to be able to set this to their required level.
> 
I dont see where the estimation happens twice. If you don't have a default CI in tidycmprsk we can add it, but then we need to include testing for that argument

> > * Do we want to add an overall strata? We can put Overall instead of 1 inside the formula and then add a column strat99= Overall
> 
> Please let's keep it as it is. It'll avoid downstream changes that would be required, and using the 1 in the formula is what `estimate_KM()` does as well.

`estimate_KM` does put 1, but changes the strata to Overall. We can change later if required.

SHAESEN2:
> > * %||% is not defined. When strata is NULL then the formula becomes ...... ~ NULL in the final object without any error => used ifelse statement
> 
> It's fine if you prefer not to use it, but it is defined and imported from {rlang}

Somewhere it goes wrong since it was not recognized during my session. Shall we then remove the import?

ddsjoberg:
@SHAESEN2 I will respond to each of your comments now. But any changes you feel need to be made to the already reviewed and approved code should be addressed separately from this pull request. Let's setup a time to video chat about the changes you want, and I think we can quickly come to a resolution.

R/estimate_cuminc.R
ddsjoberg: There are far fewer users who need competing risks estimation. Let's just keep the one example here showing the results of `estimate_cuminc()` and one visualization.

R/estimate_cuminc.R
ddsjoberg: > Based on your feedback, I assume there is no default CI estimation in tidycmprsk? 

I don't understand this comment. The CIs are estimated up front, which is why the first function needs a confidence interval argument.

ddsjoberg:
> estimate_KM does put 1, but changes the strata to Overall. We can change later if required.

Please do not update. First, this structure is used downstream in many other functions. Second, the output from `estimate_cuminc()` and `tidycmprsk::cuminc()` are identical, which is quite advantageous for many others who don't work in a CDISC env.

ddsjoberg:
> > > * %||% is not defined. When strata is NULL then the formula becomes ...... ~ NULL in the final object without any error => used ifelse statement
> > 
> > 
> > It's fine if you prefer not to use it, but it is defined and imported from {rlang}
> 
> Somewhere it goes wrong since it was not recognized during my session. Shall we then remove the import?

There is no need to remove and it's being used throughout the package.

SHAESEN2:
> > The updates look good to me. Just two comments.
> > > * focus on arguments that are required for ADaM eg confint can be defined through ... Tidycmprisk should deal with confint variations
> > 
> > 
> > Putting the `conf.int=` argument in `estimate_cuminc()` avoids an unnecessary double estimation of the entire object. Please add it back. The returned object from this function includes CIs and the users need to be able to set this to their required level.
> 
> I dont see where the estimation happens twice. If you don't have a default CI in tidycmprsk we can add it, but then we need to include testing for that argument

==> Meeting of 02MAY2022 tidycmprsk uses conf_level and therefore we need to pass conf.int earlier to the function.

> 
> > > * Do we want to add an overall strata? We can put Overall instead of 1 inside the formula and then add a column strat99= Overall
> > 
> > 
> > Please let's keep it as it is. It'll avoid downstream changes that would be required, and using the 1 in the formula is what `estimate_KM()` does as well.
> 
> `estimate_KM` does put 1, but changes the strata to Overall. We can change later if required.

R/estimate_cuminc.R
SHAESEN2: We will keep as discussed during the meeting of 2022 05 02 to keep untill users complain

PR #366, opened 2022-05-07
Replaced `survfit` call with quo

### LINKED ISSUE #304: tidyme.survfit ( call ) and dependent friends
Daniel proposed to attach the survfit object to the estimated object rather than saving the call inside the dataframe because it might cause issues when users want to use this call and are not working in the associated environment. Another solution is to work with quosures.

### LINKED ISSUE #365: Saved `survfit()` call in `estimate_KM()` discrepancy
When working with non-CDISC data (which many do), it's common that we'll pipe the dataset into a `mutate()` to make a new censoring variable, then pipe that dataset into `estimate_KM()`. Within visR, the call is modified to remove the `.` and replace it with the data set name. How is this call used downstream? Is it ever evaluated? If so, it can frequently return the wrong result.

In the example below, I subset the lung data set and recode the status variable. But the call returned from `estimate_KM()` include a call to `data = lung` (not the modified lung data object). It seems like this is likely to be an error somewhere, I think particularly in the `get_pvalue()` function?

``` r
library(tidyverse)
library(survival)

km <- 
  lung %>%
  filter(sex == 1) %>%
  mutate(status = case_when(status == 2 ~ 0, status == 1 ~ 1)) %>%
  visR::estimate_KM(
    CNSR = "status",
    AVAL = "time",
    strata = "ph.ecog"
  )
km$call
#> survival::survfit(formula = survival::Surv(time, 1 - status) ~ 
#>     ph.ecog, data = lung)
```

<sup>Created on 2022-05-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

ddsjoberg:
- The call saved in a `survift` object has been replaced with an {rlang} quo. The quo preserves the original calling environment, and this saved environment is used multiple times in downstream functions to evaluate and parse the original call to `survfit()`. (#365) 
- The name of the original data frame passed to `estimate_KM()` is now exported as `.$data_name`.

 (These items can be copied directly into `NEWS.md` when it's ready to merge)

Unit tests for `get_pvalue()` have been added as well a unit test for `.$data_name`.

closes #365
closes #304

Checklist for PR reviewer

- [ ] Approve Pull Request


ddsjoberg:
@SHAESEN2 this is pretty close to complete. There is something tricky happening with `get_COX_HR()`: the tests and examples all run interactively, but we hit an error with R CMD Check. I'll look into it again later.

ddsjoberg:
@SHAESEN2 this is now ready for you

ddsjoberg:
A couple of notes that may make review easier:
- `rlang::quo()` is similar to creating an expression, except that it also attached the associated environment as an attribute.
- `rlang::quo_squash()` converts a `quo()` object into an expression. This is used frequently to convert the `survfit()` quo into an expression so we can parse the formula or data objects from the expression or use base `eval()` (which won't work on a quo).
- `rlang::inject()`: this function is essentially `rlang::expr() %>% eval()`; it creates an expression and immediately evaluates it.
- bang-bang: The `!!` operator (or bang-bang) injects an object's value into an expression. See example below

``` r
rlang::expr(x %in% letters[1])
#> x %in% letters[1]
rlang::expr(x %in% !!letters[1])
#> x %in% "a"
```

<sup>Created on 2022-05-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>
cc @SHAESEN2

R/get_pvalue.R
```diff
@@ -74,21 +74,22 @@ get_pvalue <- function(survfit_object,
...
     ))),
     `Custom`      = rlang::expr(eval(as.call(
-      append(NewCall, list(rho = rho))
+      append(!!NewCall, list(rho = !!rho))
     )))
   )[ptype]
 
 
-  survdifflist_eval <- lapply(survdifflist, eval, env = environment())
+  survdifflist_eval <-
```
SHAESEN2, line 91: We only wish to evaluate what is required. It is to be more efficient.

R/estimate_KM.R
```diff
@@ -65,102 +65,56 @@
...
-  if (is.null(data)) stop(paste0("Data can't be NULL."))
-
-  if (base::length(base::deparse(Call[["data"]])) == 1 && base::deparse(Call[["data"]]) %in% c(".", ".x")){
-    df <- the_lhs()
-    Call[["data"]] <- as.symbol(df)
-  } else {
-     df <- as.character(sub("\\[.*$", "", deparse(dfExpr))[1])
-  }
-
-  if (!(inherits(data, "data.frame") | inherits(data, "tibble") | inherits(data, "data.table"))) stop(paste0("Data can be of `class` dataframe or tibble or data.table."))
```
SHAESEN2, line 96: Test needs to be included

R/estimate_KM.R
```diff
@@ -65,102 +65,56 @@
...
-
-  data <- as.data.frame(data)
-
-# Validate columns --------------------------------------------------------
```
SHAESEN2, line 100: keep comments (throughout document)

R/estimate_KM.R
```diff
@@ -65,102 +65,56 @@
...
+  if (!is.numeric(data[[AVAL]])){
     stop("Analysis variable (AVAL) is not numeric.")
   }
 
-  if (! is.numeric(data[[CNSR]])){
+  if (!is.numeric(data[[CNSR]])){
     stop("Censor variable (CNSR) is not numeric.")
   }
 
-# Ensure the presence of at least one strata -----------------------------
```
SHAESEN2, line 116: I prefer this explicit code section for easy debugging and understanding of what happens rather than to be more compact. Our target audience included SAS programmers that have little experience in R

SHAESEN2: If you prefer formula_rhs instead of main, that is OK for me

R/estimate_KM.R
```diff
@@ -65,102 +65,56 @@
...
 
- ## Reverse censoring: see ADaM guidelines versus R survival KM analysis
+  survfit_object <-
+    rlang::inject(survival::survfit(!!formula, data = data, !!!dots)) %>%
+    survival::survfit0(start.time = 0)
 
-  formula <- stats::as.formula(paste0("survival::Surv(", AVAL, ", 1-", CNSR, ") ~ ", main))
+  # convert survfit() call to quo with attached envir --------------------------
+  survfit_object$call[[1]] <- rlang::expr(survival::survfit) # adding `survival::` prefix
+  survfit_object$call <- rlang::quo(!!survfit_object$call)
```
SHAESEN2, line 112: for me to check if environment is also attached or needs to be specified

SHAESEN2: for me to check if ... (ellipsis) is also there

R/estimate_KM.R
```diff
@@ -194,9 +148,18 @@ estimate_KM <- function(
...
 
+  # Return ------------------------------------------------------------------
+  survfit_object
+}
 
-
-# Return ------------------------------------------------------------------
-
-  return(survfit_object)
+.call_list_to_name <- function(call_list) {
```
SHAESEN2, line 155: check overlap with the_lhs. We should only keep one.

tests/testthat/test-watchdog_testfiles.R
```diff
@@ -10,6 +10,7 @@
 
 context("validate_watchdog - T1. Details on last change for test files are recorded.")
 # skip on GH Actions and CRAN
+skip_if(tolower(Sys.info()[["user"]]) %in% "sjobergd")
```
SHAESEN2, line 13: remove

tests/testthat/test-get_pvalue.R
```diff
@@ -38,6 +38,24 @@ testthat::test_that("T1.1. No error when a `survfit` object is passed to the fun
 
   survfit_object <- visR::estimate_KM(adtte, strata = "TRTA")
   testthat::expect_error(visR::get_pvalue(survfit_object), NA)
+
+  # testing the p-value is correct when filtered data is piped to estimate_KM()
```
SHAESEN2, line 42: this should be  a new test

SHAESEN2: also do a test within estimate_KM for this

SHAESEN2: upcate cox testing too

tests/testthat/test-estimate_KM.R
```diff
@@ -380,36 +380,28 @@ testthat::test_that("T6.4 The function adds strata labels equal to the strata na
...
-
-  ## survival package
-  survobj_survival <- adtte %>%
-    survival::survfit(survival::Surv(AVAL, 1-CNSR) ~ SEX, data = .) %>%
-    survival::survfit0(start.time = 0)
-  call_survival <- as.list(survobj_survival[["call"]])
-
-  ## survival package
-  survobj_visR <- adtte %>%
+testthat::test_that("T7.1 The function updates .$data_name when magrittr pipe is used", {
```
SHAESEN2, line 383: we should do this too with adtte[1:100,] or with double pipe adtte %>% mutate() %>%

ddsjoberg:
> Nice work ! I need to take more time to try it out. I don't mind doing all the updates myself while going through the code.

Thanks @SHAESEN2 ! I'll try to go through your comments and respond before our meeting today (I think I should be able to join)

R/estimate_KM.R
ddsjoberg: all those objects (data.frame, tibble, and data.table) include class "data.frame", so I updated this to a single test. Did not remove this.

R/estimate_KM.R
ddsjoberg: this still appears in the code, now on line 77

R/estimate_KM.R
ddsjoberg: Please feel free to change as you see fit. I will note that `tidyr::drop_na(any_of(c(AVAL, CNSR, strata)))` only uses top-level, user-facing functions: no behind-the-scenes-type functions here

R/estimate_KM.R
ddsjoberg: this little function calls `the_lhs()`. I didn't add this code, just moved it into a small helper function.

R/get_pvalue.R
ddsjoberg: I didn't try to improve efficiency here, just modified existing code as needed.

R/estimate_KM.R
SHAESEN2: I agree

SHAESEN2: I agree

R/estimate_KM.R
SHAESEN2: Moved to utils file

R/estimate_KM.R
```diff
@@ -194,9 +148,18 @@ estimate_KM <- function(
       rlang::set_names(strata)
   }
 
+  # Return ------------------------------------------------------------------
```
SHAESEN2, line 151: Error in attr(data[[x]], "label") %||% x : could not find function "%||%" Can you look into this? Perhaps we need to increase rlang version?

SHAESEN2, line 151: Error in attr(data[[x]], "label") %||% x : could not find function "%||%" Can you look into this? Perhaps we need to increase rlang version?

R/estimate_KM.R
SHAESEN2, line 112: For me two issues here:

1.  sf <- adtte %>%
  dplyr::filter(TRTA == "Placebo") %>%
  estimate_KM(strata = "SEX")

  > sf
Call: ~survival::survfit(formula = survival::Surv(AVAL, 1 - CNSR) ~ 
    SEX, **data = data**)
I want it to say data = "sf"

2. With this change we are breaking the compatibility with traditional survfit objects. This we wish to avoid. Could we save the environment alone into an argument and then reconstruct the call if an environment is attached? Otherwise we redo the call as we did before?

SHAESEN2, line 112: For me two issues here:

1.  sf <- adtte %>%
  dplyr::filter(TRTA == "Placebo") %>%
  estimate_KM(strata = "SEX")

  > sf
Call: ~survival::survfit(formula = survival::Surv(AVAL, 1 - CNSR) ~ 
    SEX, **data = data**)
I want it to say data = "sf"

2. With this change we are breaking the compatibility with traditional survfit objects. This we wish to avoid. Could we save the environment alone into an argument and then reconstruct the call if an environment is attached? Otherwise we redo the call as we did before?

SHAESEN2: OK

R/estimate_KM.R
```diff
@@ -65,101 +65,68 @@
...
 
-  data <- as.data.frame(data) %>%
-    tidyr::drop_na(AVAL, CNSR)
+  data <-
+    as.data.frame(data) %>%
+    tidyr::drop_na(any_of(c(AVAL, CNSR)))
 
   if (!is.null(strata)){
     data <- data %>%
       tidyr::drop_na(any_of({{strata}}))
```
ddsjoberg, line 108: Remove the `{{ }}` here. The `strata` object can only be a character vector and you use the curly-curly when working the captured quos

R/estimate_KM.R
ddsjoberg: 1. The sf name is saved in `data_name`, not in the quo. The quo must evaluate properly and this would not happen if we changed the data object.
2. I started looking into this, and was happy to see that everything works wit the `survfit()` object as well. There was only once instance I tested that errored and that is when the survfit object is created like `data %>% filter() %>% survfit(one ~ two, data = .)`, `get_pvalue()` is later called on the object. BUT this has never worked in the package, and is a key example of when you need the correct envir specified to evaluated.

R/estimate_KM.R
ddsjoberg: It is working on my machine and on GH Actions. What notes/errors do you see when you run R CMD Check locally?

R/estimate_KM.R
SHAESEN2: The main issue is that I wish to see the correct call when printing the object. data = data is just not informative. Since the mofied dataframe is inside the same environment, I don't see why it can't be captured. Perhaps we should create a new variable with the name of the actual data 'name of original df' <- data and then create the quosure. Would that work?

SHAESEN2: The main issue is that I wish to see the correct call when printing the object. data = data is just not informative. Since the mofied dataframe is inside the same environment, I don't see why it can't be captured. Perhaps we should create a new variable with the name of the actual data 'name of original df' <- data and then create the quosure. Would that work?

SHAESEN2: For the second issue, we need to investigate further. Did you find why it does not work?

ddsjoberg: 1. expressions, quosures, and environments are already difficult to handle. The more we massage them and manipulate them, we are likely to not cover edge cases and will introduce bugs. Also, we don't have a reliable method for extracting the data set symbol, which is why we are here adding quosures and environments: our previous method led to a silent bug when trying to extract the input data set symbol and use it in `get_pvalue()`. The tilda before the quosure is how the are printed and cannot be modified. The raw output from a survfit object isn't useful anyway, so let's not complicate the situation by trying to modifying it.
2. I do know why this doesn't work. If you construct a `survfit()` using `%>%` and `data = .`, the saved call will contain `data = .`. To calculate the p-value, the call is parsed and reconstructed into a `survdiff()` call, but `data = .` can never be recovered. We could improve error messaging in this situation, but there is nothing we can do to avoid the error.

R/get_pvalue.R
ddsjoberg: Took a quick look at this. This evaluated list is used multiple times in the section below. I would need to dive deeper into the `get_pvalue()` code to see exactly what is needed and what can be removed. Let's save this for a separate pull request whose aim is to improve efficiency 

```r
 survdifflist_eval <-
    lapply(survdifflist, eval, envir = attr(survfit_object$call, ".Environment"))

# Statlist ----------------------------------------------------------------

  statlist <- unique(statlist)
  statlist <- base::sub("test", "Equality across strata", statlist, fixed = TRUE)
  statlist <- base::sub("pvalue", "p-value", statlist, fixed = TRUE)
  Nms <- names(survdifflist_eval)

  stat_summary <- list(
    `Equality across strata` = rlang::expr(base::sub(
      "Custom",
      paste0("Harrington and Fleming test (rho = ", rho, ")"),
      Nms,
      fixed = TRUE
    )),
    `Chisq`    = rlang::expr(unlist(
      lapply(survdifflist_eval, function(x)
        format(round(x$chisq, 3), nsmall = 3, justify = "right", width = 6, scientific = FALSE)
      ))),
    df        = rlang::expr(unlist(
      lapply(survdifflist_eval, function(x)
        length(x$n) - 1)
    )),
    `p-value` = rlang::expr(unlist(
      lapply(survdifflist_eval, function(x)
        .pvalformat(
          stats::pchisq(x$chisq, length(x$n) - 1, lower.tail = FALSE)
        ))
    ))
  )[statlist]
```

tests/testthat/test-get_pvalue.R
ddsjoberg: Added this test to its own section, and added similar tests in `estimate_KM()` as well. The cox model hasn't been merged yet.

tests/testthat/test-estimate_KM.R
ddsjoberg: I added a test with `adtte[1:100,]`

R/estimate_KM.R
ddsjoberg: I removed the curly brackets

R/estimate_KM.R
```diff
@@ -65,101 +65,68 @@
...
-    survfit_object, start.time = 0
-  )
+  survfit_object <-
+    rlang::inject(survival::survfit(!!formula, data = data, !!!dots)) %>% # immediate resolves call arguments
+    survival::survfit0(start.time = 0)
 
+  # convert survfit() call to quo with attached envir --------------------------
 
-# Update Call with original info and dots, similar as update.default ------
+  survfit_object$call[[1]] <- rlang::expr(survival::survfit) # adding `survival::` prefix
```
SHAESEN2, line 126: survfit_object$call[[1]] <- rlang::expr(survival::survfit) # adding `survival::` prefix
  survfit_object$call <- rlang::quo(!!survfit_object$call)

When looking into detail I see the first line is overwritten by the second line of code. One aesthetic issue that bothers me is that the quosure is creating ~ in front of the formula. Do you know why that is introduced? It will confuse less experienced programmers.

R/get_pvalue.R
SHAESEN2: OK but then put the original code back. How it is written originally, it create an unevaluated list and only evaluated the required elements.

ddsjoberg: The update I made was to evaluate in the correct environment. You'll need to explain to me how the previous code only evaluated a single call. Let's either join Mondays call early to discuss or stay on after.

R/estimate_KM.R
ddsjoberg: The first line just adds the survival prefix to the call ( which we can remove, I left it in as it previously appeared ), the second converts the call to a quosure. Is there an issue with the results?

SHAESEN2: Looking at it again, it makes more sense.

SHAESEN2:
PR is OK for me except for two issues I see:

ddsjoberg:
> PR is OK for me except for two issues I see:
> 
> 1. Below returns a character vs symbol. It would be helpful if we could return the character.
>    survobj <- visR::estimate_KM(data = adtte[adtte$SEX == "F", ], strata = "RACE")
>    testthat::expect_equal(survobj$data_name, "adtte")

I change them to character
 
> 2. There is an error in this PR
> 
> library(visR)
> 
> sf <- adtte %>% dplyr::filter(TRTA == "Placebo") %>% estimate_KM(strata = "SEX")
> 
> Error in .call_list_to_name(as.list(match.call())) : could not find function ".call_list_to_name"

This isn't an issue on my machine or on GitHub Actions.

ddsjoberg:
all checks pass except some Ubuntu builds that are failing to build R; not an issue with the code.

PR #405, opened 2022-06-12
Updates from review estimate_KM

SHAESEN2:
Updated estimate KM documentation to ensure to user knows that we allow non-CDISC. This was not clear.
Updated estimate KM testing

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -3,58 +3,19 @@ Package: visR
 Title: Clinical Graphs and Tables Adhering to Graphical Principles
 Version: 0.2.0.9011
 Authors@R: c(
-    person(given = "Mark",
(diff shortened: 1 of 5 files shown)


ddsjoberg (approved):
Changes look great. I made the following updates.
- The lifecycle package was listed twice in the DESCRIPTION file leading to a note in R CMD Checks. I removed one of the entries and ran `usethis::use_tidy_description()` to put all DESCIRPTION items in the standard format, including alphabetizing package dependencies. This should help us see if a package has already been added when adding a new dependency.
- There was an error message that was removed from `get_pvalue()`. I re-added it, but in a way that maintains 100% code coverage.
- Made slight modification to the text in the help file for estimate_KM

PR #427, opened 2022-06-24
Updated shinyapps deploy branch to main

timtreis:
Changing workflow to deploy from `main` instead of `develop`


ddsjoberg:
@timtreis change looks good to me! But tricky to test until we merge it 😆

ddsjoberg (approved):
✔️

timtreis:
@ddsjoberg feel free to merge once the OSX check completed, I think this should fix the workflow. At least that was the only error I saw in the log. I'll be on a place so I can't

PR #437, opened 2022-06-29
Zooming in on figure with `coord_cartesean()`

### LINKED ISSUE #402: KM figures cut off at ends with `visr(x_ticks=)`
When using `visr.survfit(x_ticks=)` to reduce the range of values shown, the curves a cut off early, rather than being zoomed in on. In the example below, the KM curves are cut off, but you can see from the risk table that them are still patients at risk at 10 years and the lines do not appear.

Perhaps we should use coord zoom rather than xlim?

image

ddsjoberg:
* Zooming in on figure with `coord_cartesean()` instead of using `xlim=` and `ylim=`. The latter first remove data outside the limits, then construct the line. Zooming in constructs the full line, then zooms into the limits. This is useful because the risktable often reports estimates near the end of a KM figure, but the line is cutoff and not shown at the last timepoint. (#402)

The checks ran without issue on my machine. I am not sure what unit tests can be added to check the results at the moment. I think this will be checked when we implement snapshots in testthat.

closes #402
closes #437

Here is what the update looks like:
``` r
library(visR)

estimate_KM(
  data = survival::lung %>% dplyr::mutate(time = ifelse(time > 1000, 2001, time)),
  formula = Surv(time, status) ~ 1
) %>%
  visr(x_ticks = seq(0, 2000, by = 200)) %>%
  add_risktable()
```

<sup>Created on 2022-06-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

And this is what was previously shown:

I also added the optional development mode from devtools. If any devs use dev mode, it'll be activated. If you don't, there is no change at all.

Checklist for PR reviewer

- [ ] Run `devtools::build_readme()` to build the `README.md` file.
- [ ] Approve Pull Request


SHAESEN2:
This makes sense !

SHAESEN2 (changes requested):
Please include testing for this. You can choose to use a visual compare or deconstruct the ggplot to question what is being displayed.

ddsjoberg:
I did some reading on snapshots and vdiffr. It seems that vdiffr is a special type of testthat snapshot. I added a vdiffr snapshot test for this pull request. @SHAESEN2 

I saw that the _snaps folder, however, was included in the gitignore. I don't think the snapshots can be compared on GH Actions when this folder is not included in the repo. I navigated to the ggplot2 repo, and their snaps folder is indeed uploaded to GH. So I removed the folder from the gitignore, and now all the vidffr snapshots are also checked on GH Actions.

SHAESEN2:
I think we should then also update all our testing with snapshots because most of it is out of date.

tests/testthat/test-visr_survfit.R
```diff
@@ -47,6 +47,8 @@
 #' T5. The final object is a ggplot of class `ggsurvfit`.
 #' T5.1 The final object is a ggplot of class `ggplot`.
 #' T5.2 The final object is a ggplot of class `ggsurvfit`.
+#' T6. The final object does not exclude parts of KM estimate.
+#' T6.1 The final object zooms and does not exclude trialing pieces of lines..
```
SHAESEN2, line 51: remove .. => .

tests/testthat/test-visr_survfit.R
```diff
@@ -510,4 +512,20 @@ testthat::test_that("T5.2 The final object is a ggplot of class `ggsurvfit`.", {
   testthat::expect_true(inherits(survfit_plot, "ggsurvfit"))
 })
 
+# Requirement T6 ---------------------------------------------------------------
+
+testthat::context("visr_plot - T6. The final object does not exclude parts of KM estimate.")
+
+testthat::test_that("T6.1 The final object zooms and does not exclude trialing pieces of lines.", {
```
SHAESEN2, line 519: I'm not a big fan of creating the actual and test plot with similar code. We should update this throughout the package. Could we somehow get the line information and ensure it is displayed?

ddsjoberg: I am not sure what you mean. The snap (aka the svg image of the ggplot) is created from the code and the unit test will always correspond to the code....

ddsjoberg:
@SHAESEN2 @bailliem are there updates needed before this is merged?

tests/testthat/test-visr_survfit.R
SHAESEN2: It should be independent. Perhaps we can organize a spring where we implement this. For now OK.

tests/testthat/test-visr_survfit.R
```diff
@@ -47,6 +47,8 @@
 #' T5. The final object is a ggplot of class `ggsurvfit`.
 #' T5.1 The final object is a ggplot of class `ggplot`.
 #' T5.2 The final object is a ggplot of class `ggsurvfit`.
+#' T6. The final object does not exclude parts of KM estimate.
```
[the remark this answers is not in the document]
SHAESEN2: Because it is vague, I dont know what you are testing:
_The final object does not exclude parts of KM estimate._ Which parts? and what is the purpose? Did you mean that the object will display all available time points even if the x axis argument is used?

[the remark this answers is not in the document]
ddsjoberg: This entire PR makes one small change, and from my POV it seems like a bad-faith argument that it's unclear what is intended to be tested here (even if there were no text describing it). If you would like something else here, go ahead and update it.

PR #442, opened 2022-07-13
Closing data connection in interactive vignette

timtreis:
Currently on the phone, will update later


ddsjoberg:
🤔 🤔 🤔 the test coverage still fails. I am not sure that the vignette is built when the `covr::` function is run. I think the connection error is from `test-vignette.R`? that file has explicate calls to a connection...

